PR #2, opened 2015-09-18
Add school norm

dsquirek:


SchoolNormalization.md
```diff
@@ -0,0 +1,235 @@
...
+                "confidence": 2
+            }
+        ]
+    }
+}
+```
+
+#Versioning
+-----------
+The response from the School Normalization Service is versioned with the current version being 1.0. The data set used to preform the normalization is also versioned with the current version being 1.0.0, however customers will not have the ability to use older versions of the data set as the updates are done to improve the accuracy of the results rather, and will not impact the response.
```
matthuhman: perform*

SchoolNormalization.md
```diff
@@ -0,0 +1,235 @@
...
+
+
+
+#Request Information
+
+
+HTTP method: GET or POST
+Parameters (query/form):
+-        query (required) : The educational instutution information to be queried.
+-        country (optional) : 2 digit ISO country code of eductaionl institution. Accepted country codes are [here](#supported-countries)
```
matthuhman: educational*

matthuhman:
Looks good once typos are fixed

SchoolNormalization.md
```diff
@@ -0,0 +1,235 @@
...
+- [Supported Countries](#supported-countries)
+
+
+
+#Request Information
+
+
+HTTP method: GET or POST
+Parameters (query/form):
+-        query (required) : The educational instutution information to be queried.
```
thomasmahoney: institution*

SchoolNormalization.md
```diff
@@ -0,0 +1,235 @@
...
+
+```
+{
+    "data": {
+        "normalized_schools": [
+            {
+                "normalized_school_name": "Georgia Institute of Technology",
+                "id": "53bff579e4b04710d09fa98d",
+                "country": "US",
+                "surface_form": "georgia tech",
```
thomasmahoney: I just noticed this -- do we want to be exposing surface_form in our service? it's not in P&N, and afaik that's usually something that the data scientists use internally. you can mark it transient if you don't want it to serialize.

SchoolNormalization.md
thomasmahoney: I would reword the second sentence here, I think it's a little confusing. Maybe something like...

"The data set used to perform the normalization is also versioned (currently 1.0.0). This data set is upgraded at will in order to improve accuracy of the classifier. Requests made to this service will always use the latest version of the data set."

SchoolNormalization.md
```diff
@@ -0,0 +1,234 @@
...
+
+```
+{
+    "data": {
+        "normalized_schools": [
+            {
+                "normalized_school_name": "Georgia Institute of Technology",
+                "id": "53bff579e4b04710d09fa98d",
+                "country": "US",
+                "confidence": 2
```
thomasmahoney: is confidence always an integer? what is the possible range? these are things that IMO aren't clear from the documentation. I think a small section explaining the fields would be good, and/or a different example response that shows multiple schools with varying confidence. if we guarantee results in decreasing order of confidence, we should say that too.

thefroatgt: If we aren't ordering, we should be, and it should be indicated in the documentation.

dsquirek: Adam we are ordering by confidence which is a double

thomasmahoney:
:+1:

PR #7, opened 2015-09-24
Update SchoolNormalization.md

dsquirek:


SchoolNormalization.md
```diff
@@ -16,7 +16,7 @@ _________
 HTTP method: GET or POST
 Parameters (query/form):
 -        query (required) : The educational institution information to be queried.
--        country (optional) : 2 digit ISO country code of educational institution. Accepted country codes are [here](#supported-countries)
+-        country (optional) : 2 digit ISO country code of educational institution. Accepted country codes are all supported CB countries. The full list can be found [here](#supported-countries)
```
thomasmahoney: I don't know if I agree with this rewording -- it's good that we brought our list up to parity with the Country table, but maybe we shouldn't be making this guarantee that the list on this page will always match up with that table. thoughts?

PR #9, opened 2015-10-15
Add taxonomy service to documentation

dsquirek:


matthuhman:
:+1:

TaxonomyService.md
```diff
@@ -0,0 +1,61 @@
...
+- [Versioning](#versioning)
+
+
+
+#Request Information
+
+
+HTTP method: GET
+Parameters (query/form):
+-        No Parameters
```
thomasmahoney: "No Parameters" isn't true -- we just don't take any _query_ parameters. The example should also be a valid request - what you call an example is more like a demonstration of the format. I would rewrite this like:

HTTP method: GET
Parameters: The only parameter accepted by this service is the name of the desired taxonomy, provided as part of the URL path.

```
https://api.careerbuilder.com/core/taxonomy/{taxonomyName}
```

Example: https://api.careerbuilder.com/core/taxonomy/onet17

TaxonomyService.md
```diff
@@ -0,0 +1,61 @@
...
+            }
+        ]
+    }
+}
+```
+
+
+#Response Information
+----------
+The response returns a single data node which contains a list of languages. Each language contains a list of taxonomies within that language.
```
thomasmahoney: This is not completely correct. Lists in JSON start with [, so "data" doesn't contain a list. I would maybe write it like:

The response will contain a single node `data` which maps each language code supported by the taxonomy to a list of the taxonomy's elements for that language.

TaxonomyService.md
```diff
@@ -0,0 +1,61 @@
...
+
```
thomasmahoney: also, I think you should move the response description before the example response.

TaxonomyService.md
```diff
@@ -0,0 +1,62 @@
...
+- [Available Taxonomies](#available-taxonomies)
+- [Versioning](#versioning)
+
+
+
+#Request Information
+
+
+HTTP method: GET
+Parameters: The only parameter accepted by this service is the name of the desired taxonomy, provided as part of the URL path.
```
thomasmahoney: Let's also add "Taxonomy names are case-sensitive."

TaxonomyService.md
```diff
@@ -0,0 +1,62 @@
...
+
+https://api.careerbuilder.com/core/taxonomy/{taxonomyName}
+ 
+Example: https://api.careerbuilder.com/core/taxonomy/ ([Desired Taxonomy](#available-taxonomies))/
```
thomasmahoney: You forgot to update this. make it /taxonomy/onet17 or something

TaxonomyService.md
```diff
@@ -0,0 +1,62 @@
...
+            }
+        ]
+    }
+}
+```
+
+
+#Response Information
+----------
+The response returns a single data node which maps each language code supported by the taxonomy to a list of the taxonomy's elements for that language.
```
thomasmahoney: can we move this above the sample response?

thomasmahoney:
:+1:

PR #25, opened 2016-03-02
document 17k char limit, remove pages file warning

thomasmahoney:
2 months is long enough, shouldn't need to keep that .pages warning around forever. also added a note about image files


dsquirek:
:+1:

PR #35, opened 2016-06-13
document geocoding service

thomasmahoney:


dsquirek:
might want to add a link to geocoding in the readme

PR #36, opened 2016-06-16
Add geovalidation

dsquirek:


Geovalidation.md
```diff
@@ -0,0 +1,143 @@
...
+
+- [Description](#description)
+- [Request Structure](#request-structure)
+- [Response Structure](#response-structure)
+- [Versioning](#versioning)
+
+----------------------------
+#Description
+
+Accepts structured or unstructured location information and tests the validity of the resulting resolved geographic coordinate based on the strategies requested and the location types accepted. This service supports the HTTP/GET method only.
```
thomasmahoney: should just be "structured", this service doesn't do unstructured

"and tests the validity of the resulting resolved geographic coordinate"

this is worded kind of awkwardly (and isn't exactly true, we don't validate the coordinate), how about something like:

"Accepts structured location data and attempts to geocode the data using the validation strategies specified in the request. Each successful geocoding call is verified against the list of acceptable location types specified in the request. The outcome of each strategy is described in the response, along with the outputs of the geocoding calls."

Geovalidation.md
```diff
@@ -0,0 +1,143 @@
...
+
+----------------------------
+#Request Structure
+
+Example URL: https://api.careerbuilder.com/core/geography/validation?locality=atlanta&country=US&validation_strategies=FREETEXT&accepted_location_types=locality
+
+At least one of (query, locality, postal_code, admin_area, country) must be provided in the query string, as well as a comma separated strings of valid accepted location types, and validation stratgies.
```
thomasmahoney: "as well as a comma separated strings of valid accepted location types, and validation stratgies."

-->

"as well as a "validation_strategies" parameter and an "accepted_location_types" parameter, each a comma-separated list of one or more values."

Geovalidation.md
```diff
@@ -0,0 +1,143 @@
...
+
+
+| Parameter  | Description |
+|------------|-------------|
+| address      | Optional. A string containing address information, e.g. "5550 Peachtree Pkwy."" Other forms of location data may be passed in this field (such as postal codes or country names), but we recommend using the type-specific parameters when the location type of the data is known. This parameter is handled in the same way as the **query** parameter and the two cannot be used simultaneously. |
+| query      | Optional. A string containing location information in any format, e.g. "5550 Peachtree Pkwy, Norcross, GA 30092."" We recommend using the type-specific parameters when the location type of the data is known. This parameter is handled in the same way as the **address** parameter and the two cannot be used simultaneously. |
```
thomasmahoney: this should be removed

Geovalidation.md
```diff
@@ -0,0 +1,143 @@
...
+#Request Structure
+
+Example URL: https://api.careerbuilder.com/core/geography/validation?locality=atlanta&country=US&validation_strategies=FREETEXT&accepted_location_types=locality
+
+At least one of (query, locality, postal_code, admin_area, country) must be provided in the query string, as well as a comma separated strings of valid accepted location types, and validation stratgies.
+
+
+| Parameter  | Description |
+|------------|-------------|
+| address      | Optional. A string containing address information, e.g. "5550 Peachtree Pkwy."" Other forms of location data may be passed in this field (such as postal codes or country names), but we recommend using the type-specific parameters when the location type of the data is known. This parameter is handled in the same way as the **query** parameter and the two cannot be used simultaneously. |
```
thomasmahoney: rather than copying and pasting documentation from geocoding (is this really any better than copying and pasting code?) let's change the wording to state that we accept all of the same parameters as the geocoding service, and link to that part of our documentation. we can name the parameters, but let's not duplicate the task of documenting them.

Geovalidation.md
```diff
@@ -0,0 +1,143 @@
...
+| try_locality_as_admin_area | Optional. When enabled, a request that includes the **locality** parameter and fails to retrieve locality-specific geography data will be reattempted with the locality value sent as an admin_area value instead. (If an admin_area value already exists, the locality value will be prepended and comma-separated, e.g. "locality=Harris%20Township&admin_area=OH" would be resent as "admin_area=Harris Township, OH" if try_locality_as_admin_area is set to true.) Default value is false. |
+| validationStrategies | Required. [Validation Strategies](#Validation Strategies) for making the geocoding request. |
+| acceptedLocationTypes | Required. [Location types](#Location types) to validate the geocoding response by. |
```
thomasmahoney: you got the casing wrong on these two

Geovalidation.md
```diff
@@ -0,0 +1,143 @@
...
+
+&nbsp;
+#Validation Strategies
+
+| Validation Strategy | Description|
+|----------|-------------|
+| FIELDED | Make a Geocoder call with admin area, address, country, locality, and postal code |
+| FIELDED_NO_ADDRESS | Make a Geocoder call with the provided admin area, country, locality, and postal code |
+| FREETEXT | Makes a Geocoder call with the query parameter equal to the formated string "address, locality, admin area, country, postal code"  |
+| POSTAL_CODE_ONLY |  Makes a Geocoder call with postal code parameter set to the provided postal code |
```
thomasmahoney: is this really what it does? this strategy needs to pass country too, or it's not useful

Geovalidation.md
```diff
@@ -0,0 +1,143 @@
...
+| try_locality_as_admin_area | Optional. When enabled, a request that includes the **locality** parameter and fails to retrieve locality-specific geography data will be reattempted with the locality value sent as an admin_area value instead. (If an admin_area value already exists, the locality value will be prepended and comma-separated, e.g. "locality=Harris%20Township&admin_area=OH" would be resent as "admin_area=Harris Township, OH" if try_locality_as_admin_area is set to true.) Default value is false. |
+| validationStrategies | Required. [Validation Strategies](#Validation Strategies) for making the geocoding request. |
+| acceptedLocationTypes | Required. [Location types](#Location types) to validate the geocoding response by. |
+
+&nbsp;
+#Validation Strategies
+
```
thomasmahoney, line 40: I would add a note like, "Additional strategies can be implemented as needed. Please contact DataScienceApplicationDevelopment if your use case requires a strategy that's not currently available."

Geovalidation.md
```diff
@@ -0,0 +1,143 @@
...
+| Validation Strategy | Description|
+|----------|-------------|
+| FIELDED | Make a Geocoder call with admin area, address, country, locality, and postal code |
+| FIELDED_NO_ADDRESS | Make a Geocoder call with the provided admin area, country, locality, and postal code |
+| FREETEXT | Makes a Geocoder call with the query parameter equal to the formated string "address, locality, admin area, country, postal code"  |
```
thomasmahoney: your wording here is inconsistent -- "Make a Geocoder call" x2, "Makes a Geocoder call" x2

also, formatted*

Geovalidation.md
```diff
@@ -0,0 +1,143 @@
...
+| POSTAL_CODE_ONLY |  Makes a Geocoder call with postal code parameter set to the provided postal code |
+&nbsp;
+
+&nbsp;
+#Location Types
+
```
thomasmahoney, line 53: I would add a note like, "For a description of each location type, see our geocoding documentation here [link]"

Geovalidation.md
```diff
@@ -0,0 +1,143 @@
...
+&nbsp;
+
+----------------------
+#Response Structure
+
+All responses with an HTTP status of 200 will consist of a JSON object with a top-level "data" node containing the following elements:
+
+| Field    | Description |
+|----------|-------------|
+| strategy_results  | A JSON array of one or more objects containing the results of the requested strategies. The strategy_results will be ordered by their order in the validation_strategies comma separated string (ie for the string "FREETEXT,FIELDED" the result of FREETEXT will be the first result in the array) |
```
thomasmahoney, line 80: so for someone coding against this API, what is the intended usage for retrieving a given strategy result? is it just taking results[0], results[1], etc? That doesn't feel very good to me. If we enforced ALL_CAPS casing in the request, then we could guarantee that the returned strategies would have the exact same names as the input strategies. Maybe our wrappers should expose a Response.GetResult(Strategy strategy) method to make the response easier to work with?

Geovalidation.md
```diff
@@ -0,0 +1,143 @@
...
+
+&nbsp;
+
+Each element of the returned **strategy_results** array will be formatted as follows:
+
+|   |   |
+|---|---|
+| strategy | The name of the strategy. |
+| validation_data | A JSON object containing validation data. See below for properties of this object. **This object is always present regardless of the strategy result** |
+| geography | A JSON object containding the result of the geocoding call that was validated. **This object is only present if that strategy's geocoder call succeded ** |
```
thomasmahoney: containing\* and succeeded*

Geovalidation.md
```diff
@@ -0,0 +1,143 @@
...
+| expiration_date | A date string in YYYY-MM-DD format representing the last day that this data may be used. Per our contract with Google, any data cached by users of the geocoding service must be discarded by this date. |
+
+&nbsp;
+
+Properties of the **validation_data** object will contain the following properties:
+
+|   |   |
+|---|---|
+| status | The status of the valdation strategy can be (NOT_RUN, SUCCESS, FAILURE, NO_RESULTS) **This property is always present and non-empty.** |
+| message | The unabbreviated form of the administrative area's name. **This property is not always present.** |
```
thomasmahoney: what?? bad copy/paste job..?

Geovalidation.md
```diff
@@ -0,0 +1,143 @@
...
+
+&nbsp;
+
+&nbsp;
+
+Properties of the **geography** object will be formatted as follows:
```
thomasmahoney, line 106: again, just link to geocoding docs, copying and pasting is bad

Geovalidation.md
```diff
@@ -12,39 +12,40 @@
...
 
 | Validation Strategy | Description|
 |----------|-------------|
-| FIELDED | Make a Geocoder call with admin area, address, country, locality, and postal code |
-| FIELDED_NO_ADDRESS | Make a Geocoder call with the provided admin area, country, locality, and postal code |
-| FREETEXT | Makes a Geocoder call with the query parameter equal to the formated string "address, locality, admin area, country, postal code"  |
-| POSTAL_CODE_ONLY |  Makes a Geocoder call with postal code parameter set to the provided postal code |
+| FIELDED | Makes a Geocoder call with admin area, address, country, locality, and postal code |
+| FIELDED_NO_ADDRESS | Makes a Geocoder call with the provided admin area, country, locality, and postal code |
+| FREETEXT | Makes a Geocoder call with the query parameter equal to the formatted string "address, locality, admin area, country, postal code"  |
```
thomasmahoney: "with the query parameter" is more implementation detail than we should be sharing here -- I'd rephrase as "Makes a Geocoder call with a free-text string formatted as follows: [...]"

Geovalidation.md
```diff
@@ -94,7 +97,7 @@ Properties of the **validation_data** object will contain the following properti
 |   |   |
 |---|---|
 | status | The status of the valdation strategy can be (NOT_RUN, SUCCESS, FAILURE, NO_RESULTS) **This property is always present and non-empty.** |
-| message | The unabbreviated form of the administrative area's name. **This property is not always present.** |
+| message | A short message describing a reason for the returned status. **This property is not always present.** |
```
thomasmahoney: could we give more info on that last part? is it only present for non-success responses, or what?

PR #49, opened 2016-10-05
Update Geocoding.md

msundlof:


Geocoding.md
```diff
@@ -5,25 +5,25 @@
...
-
-#Place Id Requests
-
-The service also accepts Google Place Id's as a means to retrieve goeographic information. Requesting a location using a Google Place Id requires that a request only contains the Place Id and a culture param if language specification is desired. Additional location parameters will result in a 400 response in this request.
+This service makes use of two different requests: Geocoding Requests and Google Place ID Lookup. It accepts structured or unstructured location information and attempts to resolve the input to a geographic coordinate. If successful, the response will contain one (or more, if requested) locations with latitude and longitude information as well as any available administrative location information, such as country and state. This service relies on Google's Geocoding API as its data vendor; Google's service is documented [here](https://developers.google.com/maps/documentation/geocoding/intro). This service supports the HTTP/GET method only.
```
thomasmahoney: The second and third sentences in this paragraph have fallen out of date with the intro of place_id lookups. (The second is obvious; the third is because place_id only ever returns one result.) I would replace those sentences with something like: "The two request types represent two different methods of obtaining geographic information for a query (latitude/longitude, locality, administrative areas, country, etc.) and are explained in more detail below."

Oh, and ever since geo was blankspaced, the very last sentence no longer holds true. Could you change that to something like: "This service supports both HTTP GET and POST methods. Requests will be handled the same way regardless of HTTP method; use whichever method you find most convenient."

(I add that second sentence here because POST typically implies different behavior than GET. Our services typically respond to both methods in the same way, which is a deviation from standard RESTful API standards.)

Geocoding.md
```diff
@@ -34,11 +34,25 @@ At least one of (query, address, locality, postal_code, admin_area, country, pla
...
 | try_locality_as_admin_area | Optional. When enabled, a request that includes the **locality** parameter and fails to retrieve locality-specific geography data will be reattempted with the locality value sent as an admin_area value instead. (If an admin_area value already exists, the locality value will be prepended and comma-separated, e.g. "locality=Harris%20Township&admin_area=OH" would be resent as "admin_area=Harris Township, OH" if try_locality_as_admin_area is set to true.) Default value is false. |
 
+----------------------
+
+#Place ID Lookup Request
+
+A Google Place ID is a unique identifier used by Google to identify location/places in the Google Places Databases and on Google Maps. The service accepts a valid place ID as a means to retrieve goeographic information back from google via reverse geocoding. Requesting a location using a Google place ID requires that the request contains a place ID and a culture param if a language specification is desired. Additional location parameters will result in a 400 response.
```
thomasmahoney: You still have the typo on "geographic", and please capitalize Google in that sentence and lowercase "place" in the beginning of the first sentence

thomasmahoney:
:+1:

PR #55, opened 2017-01-04
Update SchoolNormalization.md

msundlof:


SchoolNormalization.md
```diff
@@ -41,7 +45,7 @@ Example: https://api.careerbuilder.com/core/normalizedschools?query=georgia%20te
 
 #Response Information
 
-The response returns a single data node which contains a list of normalized schools. These normalized schools are ordered by the confidence score. Each normalized shool has a normalized school name (string), a unique ID (string), a country (string), and a confidence (double). Confidence scores range from 0.0 to 2.0.
+The response returns a single data node which contains a list of normalized schools. These normalized schools are ordered by the confidence score. Each normalized shool has a normalized school name (string), a unique ID (string), a confidence (double), a country (string), a city (string), a state (string), an ipeds_id (string), and an ipeds_name (string). Confidence scores range from 0.0 to 2.0.
```
thomasmahoney, line 48: Maybe add a sentence here that explains what IPEDS is?

msundlof: Yeah i don't know what those are myself need to send an email heh...

gliu6: I guess it is  [Integrated Postsecondary Education Data System (IPEDS) ](https://en.wikipedia.org/wiki/Integrated_Postsecondary_Education_Data_System)

SchoolNormalization.md
```diff
@@ -41,8 +45,9 @@ Example: https://api.careerbuilder.com/core/normalizedschools?query=georgia%20te
...
 
+Regarding the two fields ipeds_id and ipeds_name; ipeds stands for Integrated Postsecondary Education Data System. The unique Id and Name are assigned to every school registered with the National Center for Education Statistics.
```
gliu6: I would use a "," instead of ";". pardon me if I am wrong.
Besides I prefer "unique id and name" instead of "unique Id and Name", or just use "unique ipeds_id and ipeds_name".

thomasmahoney: Yeah, I would use either a comma or (preferably) a colon or dash; the semicolon doesn't work here. Agreed with @gliu6 on the capitalization comment too. (Btw, do we know for a fact that all IPEDS names are unique?)

msundlof: alrighty will make changes. Indeed they are unique to every school.

thomasmahoney:
:+1:

gliu6 (approved):
👍

SchoolNormalization.md
```diff
@@ -30,8 +30,12 @@ Example: https://api.careerbuilder.com/core/normalizedschools?query=georgia%20te
             {
                 "normalized_school_name": "Georgia Institute of Technology",
                 "id": "53bff579e4b04710d09fa98d",
+                "confidence": 2.0,
```
joshchao39, line 33: @msundlof Actually the confidence ranges from 0 to 1 in v2. v1 has a weird scoring system that gives score from 0 to 1 and gives score of 2 when it matches some condition.
We might want to correct this in case it confuses the customer.

PR #56, opened 2017-01-05
Create GeoEntity.md

gliu6:
The birth of GeoEntity.md

### DIFF
--- a/GeoEntity.md
+++ b/GeoEntity.md
@@ -6,8 +6,8 @@
 
 - [Description](#description)
 - [All Countries Request](#all-countries-request)
-- [All Administrative Area Ones Request](#all-administrative-area-ones-request)
-- [All Administrative Area Twos Request](#all-administrative-area-twos-request)
+- [All First-level Administrative Areas Request](#all-first-level-administrative-areas-request)
+- [All Second-level Administrative Areas Request](#all-second-level-administrative-areas-request)
 - [All Localities Request](#all-localities-request)
 - [All Postal Codes Request](#all-postal-codes-request)
 - [All Metropolitan Statistical Areas Request](#all-metropolitan-statistical-areas-request)
@@ -17,10 +17,10 @@
 ----------------------------
 #Description
 
-This service provides aggregated geography entities. Currently the following requests could be made.
+This service provides lists of geographic entities of a particular type, such as postal codes or first-level administrative areas. This service uses the same underlying data as our geocoding service does. The list of entities returned as quried should be comprehensive for the populous countries, but we could not guarantee the completeness. We are considering to provide features to add missing entities if there are any and as customers request. Currently, the following requests are supported:
 - Request all countries in the world.
 - Request all administrative area ones in a specific country.
-- In a specific country ( and a specific first order of administrative area as optional), 
+- In a specific country (optionally limited to a single first-level administrative area):
 	- Request all administrative area twos.
 	- Request all localities.
 	- Request all postal codes.
@@ -33,9 +33,9 @@ All responses with an HTTP status of 200 will consist of a JSON object with a to
 ----------------------------
 #All Countries Request
 - Request Structure:
-	- Target: http://localhost:8080/admin_area_1s (After adding to Routing layer, we update this)
+	- Target: https://api.careerbuilder.com/core/geography/entities/countries 
 	- URL parameter: none
-	- Example URL:http://localhost:8080/countries
+	- Example URL:https://api.careerbuilder.com/core/geography/entities/countries
 	
 - Response Sample:
 ```
@@ -50,7 +50,6 @@ All responses with an HTTP status of 200 will consist of a JSON object with a to
           "lat": 33.93911,
           "lng": 67.709953
         },
-        "location_type": "APPROXIMATE",
         "viewport": {
           "northeast": {
             "lat": 38.4908389,
@@ -62,23 +61,24 @@ All responses with an HTTP status of 200 will consist of a JSON object with a to
           }
         },
         "formatted_address": "Afghanistan"
-     }
+     },
+     ...
   ]
  }
 }
 ```
 &nbsp;
 ----------------------------
-#All Administrative Area Ones Request
+#All First-level Administrative Areas Request
 - Request Structure:
-	- Target: http://localhost:8080/admin_area_1s (After adding to Routing layer, we update this)
+	- Target: https://api.careerbuilder.com/core/geography/entities/admin_area_1s
 	- URL parameter: 
   
     | Parameter  | Required | Description |
     |----------------|----------------|-----------------|
     | country        | Yes | A country name or two-letter ISO-3166 country code, not casing sensitive            | 
     
-	- Example URL: http://localhost:8080/admin_area_1s?country=us
+	- Example URL:https://api.careerbuilder.com/core/geography/entities/admin_area_1s?country=us
 	
 - Response Sample:
 ```
@@ -94,8 +94,7 @@ All responses with an HTTP status of 200 will consist of a JSON object with a to
           "lat": 64.2008413,
           "lng": -149.4936733
         },
-        "location_type": "APPROXIMATE",
-        "viewport": {
+	"viewport": {
           "northeast": {
             "lat": 71.347066,
             "lng": -129.9945563
@@ -106,24 +105,25 @@ All responses with an HTTP status of 200 will consist of a JSON object with a to
           }
         },
         "formatted_address": "Alaska, USA"
-      }
+      },
+      ...
     ]
  }
 }
 ```
 &nbsp;
 ----------------------------
-#All Administrative Area Twos Request
+#All Second-level Administrative Areas Request
 - Request Structure:
-	- Target: http://localhost:8080/admin_area_2s (After adding to Routing layer, we update this)
+	- Target: https://api.careerbuilder.com/core/geography/entities/admin_area_2s
 	- URL parameter: 
   
     | Parameter  | Required | Description |
     |----------------|----------------|-----------------|
     | country        | Yes | A country name or two-letter ISO-3166 country code, not casing sensitive            | 
-    | admin_area_one | Optional  | A first order administrative subdivision, both long name and short name are accepted|
+    | admin_area_one | Optional  | A first order administrative subdivision, both long name and short name are accepted. If present, limits the result set to the second-level administrative areas within the specified first-level administrative area. Otherwise, all second-level administrative areas within the specified country will be returned.|
     
-	- Example URL: http://localhost:8080/admin_area_2s?country=us&admin_area_one=ga
+	- Example URL: https://api.careerbuilder.com/core/geography/entities/admin_area_2s?country=us&admin_area_one=ga
 	
 - Response Sample:
 ```
@@ -139,8 +139,7 @@ All responses with an HTTP status of 200 will consist of a JSON object with a to
           "lat": 30.898276,
           "lng": -81.6035062
         },
-        "location_type": "APPROXIMATE",
-        "viewport": {
+	"viewport": {
           "northeast": {
             "lat": 31.169595,
             "lng": -81.40354649999999
@@ -151,7 +150,8 @@ All responses with an HTTP status of 200 will consist of a JSON object with a to
           }
         },
         "formatted_address": "Camden County, GA, USA"
-      }
+      },
+      ...
   ]
  }
 }
@@ -160,15 +160,15 @@ All responses with an HTTP status of 200 will consist of a JSON object with a to
 ----------------------------
 #All Localities Request
 - Request Structure:
-	- Target: http://localhost:8080/localities (After adding to Routing layer, we update this)
+	- Target: https://api.careerbuilder.com/core/geography/entities/localities
 	- URL parameter: 
   
     | Parameter  | Required | Description |
     |----------------|----------------|-----------------|
     | country        | Yes | A country name or two-letter ISO-3166 country code, not casing sensitive            | 
-    | admin_area_one | Optional  | A first order administrative subdivision, both long name and short name are accepted|
+    | admin_area_one | Optional  | A first order administrative subdivision, both long name and short name are accepted. If present, limits the result set to the localities within the specified first-level administrative area. Otherwise, all localities within the specified country will be returned.|
     
-	- Example URL: http://localhost:8080/localities?country=us&admin_area_one=ga
+	- Example URL: https://api.careerbuilder.com/core/geography/entities/localities?country=us&admin_area_one=ga
 	
 - Response Sample:
 ```
@@ -182,7 +182,6 @@ All responses with an HTTP status of 200 will consist of a JSON object with a to
           "lat": 33.7254622,
           "lng": -84.2752831
         },
-        "location_type": "APPROXIMATE",
         "viewport": {
           "northeast": {
             "lat": 33.7399031,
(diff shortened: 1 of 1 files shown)


GeoEntity.md
```diff
@@ -0,0 +1,280 @@
...
+	- Request all metropolitan statistical areas.
+	
+This service supports both HTTP/GET and POST methods.
+All responses with an HTTP status of 200 will consist of a JSON object with a top-level "data" node containing a node "entities", which is an list of entities as requested.
+
+&nbsp;
+----------------------------
+#All Countries Request
+- Request Structure:
+	- Target: http://localhost:8080/admin_area_1s (After adding to Routing layer, we update this)
```
thomasmahoney: just go ahead and put the correct URL now. https://api.careerbuilder.com/core/geography/entities/admin_area_1s

we can wait to merge this till it's officially launched.

GeoEntity.md
```diff
@@ -0,0 +1,280 @@
...
+This service provides aggregated geography entities. Currently the following requests could be made.
+- Request all countries in the world.
+- Request all administrative area ones in a specific country.
+- In a specific country ( and a specific first order of administrative area as optional), 
+	- Request all administrative area twos.
+	- Request all localities.
+	- Request all postal codes.
+	- Request all metropolitan statistical areas.
+	
+This service supports both HTTP/GET and POST methods.
```
thomasmahoney: this service should definitely support GET only. POST does not make sense here. I think blankspace will let you do this currently?

GeoEntity.md
```diff
@@ -0,0 +1,280 @@
...
+- [All Localities Request](#all-localities-request)
+- [All Postal Codes Request](#all-postal-codes-request)
+- [All Metropolitan Statistical Areas Request](#all-metropolitan-statistical-areas-request)
+- [Versioning](#versioning)
+
+&nbsp;
+----------------------------
+#Description
+
+This service provides aggregated geography entities. Currently the following requests could be made.
```
thomasmahoney: I would rephrase these sentences a bit. maybe something like:

"This service provides complete lists of geographic entities of a particular type, such as postal codes or first-level administrative areas. Currently, the following requests are supported:"

GeoEntity.md
```diff
@@ -0,0 +1,280 @@
...
+- Request all countries in the world.
+- Request all administrative area ones in a specific country.
```
thomasmahoney, line 22: This one should be under your header on the next line, since it's country-specific. Also, I would rephrase these as "first-level administrative areas" etc instead of "administrative area ones"/"twos"

GeoEntity.md
```diff
@@ -0,0 +1,280 @@
...
+- In a specific country ( and a specific first order of administrative area as optional),
```
thomasmahoney: "(optionally limited to a single first-level administrative area):"

GeoEntity.md
thomasmahoney: nvm, I see now that your next header is only for things where you can specify an AA1.

GeoEntity.md
```diff
@@ -0,0 +1,280 @@
...
+#Description
+
+This service provides aggregated geography entities. Currently the following requests could be made.
+- Request all countries in the world.
+- Request all administrative area ones in a specific country.
+- In a specific country ( and a specific first order of administrative area as optional), 
+	- Request all administrative area twos.
+	- Request all localities.
+	- Request all postal codes.
+	- Request all metropolitan statistical areas.
```
thomasmahoney, line 27: did we never end up adding DMAs? that's ok, we can wait and see if anyone wants it.

GeoEntity.md
thomasmahoney: also, why do we have "admin_area_1s" in this section? bad copy/paste?

GeoEntity.md
```diff
@@ -0,0 +1,280 @@
...
+    "entities": [
+      {
+        "name": "Afghanistan",
+        "country_code": "AF",
+        "place_id": "ChIJbQL_-LZu0TgReNqWvg1GtfM",
+        "location": {
+          "lat": 33.93911,
+          "lng": 67.709953
+        },
+        "location_type": "APPROXIMATE",
```
thomasmahoney: let's remove Google's location_type field from the Entities API. we don't use it in geocoding (we generate our own "location_type" based on which fields are present), and it wouldn't make sense to specify a location type in this API because you're only requesting one specific entity type at a time anyway.

GeoEntity.md
```diff
@@ -0,0 +1,280 @@
...
+	
+- Response Sample:
+```
+{
+  "data": {
+    "entities": [
+      {
+        "long_name": "Alaska",
+        "short_name": "AK",
+        "level": 1,
```
thomasmahoney, line 93: no point in having this field in the API either -- as the caller, I already know that every entity returned is an AA1.

GeoEntity.md
```diff
@@ -0,0 +1,280 @@
...
+----------------------------
+#All Administrative Area Twos Request
+- Request Structure:
+	- Target: http://localhost:8080/admin_area_2s (After adding to Routing layer, we update this)
+	- URL parameter: 
+  
+    | Parameter  | Required | Description |
+    |----------------|----------------|-----------------|
+    | country        | Yes | A country name or two-letter ISO-3166 country code, not casing sensitive            | 
+    | admin_area_one | Optional  | A first order administrative subdivision, both long name and short name are accepted|
```
thomasmahoney: it might be a lot easier to just accept codes in these fields (for all endpoints, not just this one). otherwise we are vulnerable to complaints about no spelling corrections, etc.

GeoEntity.md
```diff
@@ -0,0 +1,280 @@
...
+----------------------------
+#All Localities Request
+- Request Structure:
+	- Target: http://localhost:8080/localities (After adding to Routing layer, we update this)
+	- URL parameter: 
+  
+    | Parameter  | Required | Description |
+    |----------------|----------------|-----------------|
+    | country        | Yes | A country name or two-letter ISO-3166 country code, not casing sensitive            | 
+    | admin_area_one | Optional  | A first order administrative subdivision, both long name and short name are accepted|
```
thomasmahoney: if I omit AA1, will I get e.g. all localities in the US? we should state this in the docs (for this and other endpoints). "If present, limits the result set to localities within the specified first-level administrative area. Otherwise, all localities within the specified country will be returned."

GeoEntity.md
```diff
@@ -0,0 +1,280 @@
...
+  "data": {
+    "entities": [
+      {
+        "title": "Albany, GA",
+        "code": 10500
+      },
+      {
+        "title": "Athens-Clarke County, GA",
+        "code": 12020
+      }
```
thomasmahoney: I would write this line like:
```
}, [...]
```
just to say "and so on...".

GeoEntity.md
thomasmahoney: also, you should put a bit more info in here about this service. mention that it uses the same underlying data as our geocoder. mention that we cannot guarantee 100% completeness, but the data should be fairly comprehensive. (maybe mention that we can add missing locations as needed.)

GeoEntity.md
gliu6: I went through the email thread and did not see people mentioned it.

GeoEntity.md
gliu6: Yes, it is a free feature that BS enables automatically.

thomasmahoney: no, I mean BS will let you *disable* POSTs. POSTing to this service is especially odd because it's just a static data retrieval service. I would rather it only support GETs.

GeoEntity.md
gliu6: ^

GeoEntity.md
gliu6: make sense.

GeoEntity.md
gliu6: But I guess it does not hurt. I just used one AdminArea class in several different places that they are serialzied/deserialized. It would look much more duplicate in code with the change.

GeoEntity.md
gliu6: hmm, I think it is nicer to provide both country names and code.
Imagine that customers only have a feed of country names. 
Also, i think most requests should be machine made requests, at least from our most direct customers. 
Besides, even without corrections, we could accept things like "UniTED STAtes".

GeoEntity.md
gliu6: Yes, Sure.

GeoEntity.md
gliu6: I thought about that, too. Since you have the same though, will do.

GeoEntity.md
```diff
@@ -0,0 +1,281 @@
...
+- Request all countries in the world.
+- Request all administrative area ones in a specific country.
+- In a specific country (optionally limited to a single first-level administrative area):
+	- Request all administrative area twos.
+	- Request all localities.
+	- Request all postal codes.
+	- Request all metropolitan statistical areas.
+	
+This service supports both HTTP/GET and POST methods.
+All responses with an HTTP status of 200 will consist of a JSON object with a top-level "data" node containing a node "entities", which is an list of entities as requested.
```
msundlof: typo, "which is a list" not "an list"

GeoEntity.md
```diff
@@ -0,0 +1,281 @@
...
+----------------------------
+#All Localities Request
+- Request Structure:
+	- Target: https://api.careerbuilder.com/core/geography/entities/localities
+	- URL parameter: 
+  
+    | Parameter  | Required | Description |
+    |----------------|----------------|-----------------|
+    | country        | Yes | A country name or two-letter ISO-3166 country code, not casing sensitive            | 
+    | admin_area_one | Optional  | A first order administrative subdivision, both long name and short name are accepted. If present, limits the result set to the localities within the specified first-level administrative area. Otherwise, all localities within the specified country will be returned.|
```
msundlof, line 171: feel like you need a pronoun after If present, i.e. If present, "this" or "it" limits the result... or you could just a proper noun, if present, the param limits

GeoEntity.md
```diff
@@ -0,0 +1,281 @@
...
+----------------------------
+#All Postal Codes Request
+- Request Structure:
+	- Target: https://api.careerbuilder.com/core/geography/entities/postal_codes
+	- URL parameter: 
+  
+    | Parameter  | Required | Description |
+    |----------------|----------------|-----------------|
+    | country        | Yes | A country name or two-letter ISO-3166 country code, not casing sensitive            | 
+    | admin_area_one | Optional  | A first order administrative subdivision, both long name and short name are accepted. If present, limits the result set to the postal codes ithin the specified first-level administrative area. Otherwise, all postal codes within the specified country will be returned.|
```
msundlof, line 214: same here

GeoEntity.md
```diff
@@ -0,0 +1,281 @@
...
+----------------------------
+#All Metropolitan Statistical Areas Request
+- Request Structure:
+	- Target: https://api.careerbuilder.com/core/geography/entities/msas
+	- URL parameter: 
+  
+    | Parameter  | Required | Description |
+    |----------------|----------------|-----------------|
+    | country        | Yes | A country name or two-letter ISO-3166 country code, not casing sensitive            | 
+    | admin_area_one | Optional  | A first order administrative subdivision, both long name and short name are accepted. If present, limits the result set to the metropolitan statistical areas within the specified first-level administrative area. Otherwise, all metropolitan statistical areas within the specified country will be returned.|
```
msundlof, line 257: sames

GeoEntity.md
thomasmahoney: actually, I think in this context it flows well the way it's written @msundlof, I've written similar sentences in my own documentation before. the first sentence is structured similarly (technically it's a sentence fragment, but it works here). I don't know enough about grammar to say exactly *why* all this is, but I think in the context of technical documentation this style makes sense :)

GeoEntity.md
thomasmahoney: actually, @msundlof's comment helped me notice a typo here: "ithin the specified"

GeoEntity.md
```diff
@@ -17,10 +17,10 @@
 ----------------------------
 #Description
 
-This service provides aggregated geography entities. Currently the following requests could be made.
+This service provides lists of geographic entities of a particular type, such as postal codes or first-level administrative areas. This service uses the same underlying data as our geocoding service does. The list of entities returned as quried should be comprehensive for the populous countries, but we could not guarantee the completeness. We are considering to provide features to add missing entities if there are any and as customers request. Currently, the following requests are supported:
```
thomasmahoney: "as our geocoding service does" --> either "as our geocoding service" or "that our geocoding service does"

typo in the next sentence: "quried"

"but we could not guarantee the completeness" --> "but we cannot guarantee 100% completeness" ("could" implies conditionality here, as in "we could not guarantee" [...if so-and-so were true])

"We are considering... as customers request" --> I would rephrase this sentence and move it immediately after the list of requests, before the GET/POST line (and don't forget, please kill the POST support on this service). maybe something like "If you require lists of a geographic entity type that is not on this list, please contact [team email]"

GeoEntity.md
```diff
@@ -94,8 +94,7 @@ All responses with an HTTP status of 200 will consist of a JSON object with a to
           "lat": 64.2008413,
           "lng": -149.4936733
         },
-        "location_type": "APPROXIMATE",
-        "viewport": {
+	"viewport": {
```
thomasmahoney: looks like your indentation got messed up slightly on this line. (and on other sample jsons too)

GeoEntity.md
gliu6: Glad I could improve my written English together with my coding skills.

GeoEntity.md
gliu6: yeah, I believe it is github's fault. I will check it after merging the pr.

GeoEntity.md
gliu6: @thomasmahoney what is the typo?

thomasmahoney: "**ithin** the specified"

should be "within"

for some reason, even when I wrote and reread my comment, I found it hard to catch in just the three-word quote :)

gliu6: Sorry, I still don't get it yet.
So you recommend me to change from "within the specified" to "within"?
or you mean I have lacked a letter "w" in word "within" (I don't see `w` missing)?

GeoEntity.md
gliu6: Figured out it is tab&space stuff, fixed it.

GeoEntity.md
thomasmahoney: @gliu6 bumping this "outdated" comment as some of my suggestions still apply.

thomasmahoney:
sorry @gliu6, seems like my comment keeps getting eaten. there are still some suggestions in this comment that I would like to see before I give my +1:
https://github.com/careerbuilder/DataScienceAPIDocumentation/pull/56#discussion_r95705846

gliu6:
@thomasmahoney updated. This pr behaves weird to me. I think at some point, I removed the word of "both" from "both HTTP GET" sentence and made a pr, somehow it failed.
Anyway, it is updated for sure this time.

GeoEntity.md
```diff
@@ -28,7 +28,7 @@ This service provides lists of geographic entities of a particular type, such as
 	
 If you require a list of geographic entities of a particular entity type, but the entity list is not complete, please contact <DataScienceApplicationDevelopment@careerbuilder.com>.
 
-This service supports both HTTP/GET method.
+This service supports HTTP/GET method.
```
thomasmahoney: let's make this "supports the HTTP/GET method only." then :+1: :)

PR #60, opened 2017-01-14
add remarks section to describe geo search recs

thomasmahoney:


Geocoding.md
```diff
@@ -103,6 +104,23 @@ Each element of the returned **admin_areas** array will be formatted as follows:
 &nbsp;
 
 -----------
+#Remarks and Recommendations
+
+One common use case of the geocoding API is in normalizing location queries for search purposes. Occasionally, such applications may have the design goal of performing a "contained-in" search for some entities, such as states and countries ("return all results in Texas"), and for other entities, performing a radius search ("return all results within N mi/km of this point). If this describes your use case, the following logic is recommended:
+
```
gliu6, line 110: After reading, I don't get the different between "contained-in" and radius search. To me, it seems like they are the same, just in different level and range.
Mind to explain a bit?

gliu6:
+1.

PR #65, opened 2017-01-24
Jpan

glpayson:


JobParseAndNormalize.md
```diff
@@ -0,0 +1,353 @@
...
+The Job Parse and Normalize (JPAN) service parses a Base64-encoded job posting and further enriches the parsed data with the following normalizations and classifications.  
+
+ #####Enrichments:
+ - Job Title Classifications
+    - [Carotene](https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/JobTitle.md) <sup>(optional)</sup>
+    - [ONET](https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/JobTitle.md) <sup>(optional)</sup>
+    - Textkernel
+ - [Normalized Company Name](https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/CompanyNormalization.md)  <sup>(optional)</sup>
+ - Skills Extraction
+    - [Skills 4.0](https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/Skills.md) <sup>(optional)</sup>
```
thomasmahoney: not sure we should bother with naming versions, as this will change over time (and probably fall out of date when we forget to update this). the important part IMO is "the job parsing API extracts skills."

JobParseAndNormalize.md
```diff
@@ -0,0 +1,353 @@
...
+    - Textkernel Skills
+ - Language Skills
+ - [Company Geography](https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/Geography.md) <sup>(optional)</sup>
+ - [Job Geography](https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/Geography.md) <sup>(optional)</sup>
```
thomasmahoney: Geocoding.md, not Geography. And I would probably format this a little differently - I think right now it looks like "Company Geography" and "Job Geography" are two APIs - maybe something like
`[Geocoding](link) (optional; for both company and job locations)`

JobParseAndNormalize.md
```diff
@@ -0,0 +1,353 @@
...
+
+The service is located at https://api.careerbuilder.com/core/parsing/normalizedjob. As usual, you will need OAuth core credentials to use this service. (*If you do not have these, please go [here](http://apitester.cbplatform.link/credentials) or email PlatformSoftware@careerbuilder.com to request core credentials.*)
+
+##Request Structure
+
+This service supports the HTTP GET and POST methods. Because Base64-encoded documents can be quite large, POST is encouraged for production use.
+
+The following parameters may be supplied in the query string (for HTTP GET) or form body (for HTTP POST):
+
+* **document** (Required) -- A .doc, .docx, .pdf, .rtf, .txt, .odt, .wps, and .pages documents given in a BASE64 encoded string.  Please note that the .pages format is not accepted by Textkernel; you will need to specify another parser.
```
thomasmahoney: documents --> document. And that .pages line should definitely go (both here and from list in the preceding sentence)

JobParseAndNormalize.md
```diff
@@ -0,0 +1,353 @@
...
+
+* **desired_enrichments** (Required) -- A comma-separated list, without spaces, of the desired normalization calls to perform on the results of the job parsing operation. 
+
+    This list of possible values for **desired_enrichments** and the enrichments the correspond to are as follows:
```
thomasmahoney: "enrichments **the** correspond to" --> "they"

JobParseAndNormalize.md
```diff
@@ -0,0 +1,353 @@
...
+    
+    |desired_enrichments | [Enrichment(s)](#Enrichments) |
+    | --- | --- |
+    | `company_norm` | Normalized Company Name |
+    | `geocoding` | Company Geography, Job Geography |
```
thomasmahoney, line 52: oh, you might not need what I wrote above then, since you're explaining here that geocoding enrichment entails both company and job locations. I might still rephrase this a bit as "Company location and job location," I think that's a bit clearer and the "Geography" part is already implied

JobParseAndNormalize.md
```diff
@@ -0,0 +1,353 @@
...
+    | `job_level` | Job Level |
+    | `job_title_carotene` | Carotene Job Title Classifications |
+    | `job_title_onet` | ONET Job Title Classifications |
+    | `skills` | Skills 4.0 |
```
thomasmahoney: ditto what I said above w.r.t. version

JobParseAndNormalize.md
```diff
@@ -0,0 +1,353 @@
...
+
+    For example, a request with a desired_enrichments value equal to `job_level,skills,job_title_onet,company_norm` would receive job level classifications, skills extractions, ONet job title classifications, and company normalizations. The API does not currently allow callers to request only certain versions of a classification service.
+    
+    The value "none" may be supplied to return none of the optional enrichments. At present, a request that does not include this parameter will receive all classifications; this is temporary behavior for the sake of backwards compatibility. Once all customers have started using the desired_enrichments parameter, its usage will become required, and requests excluding it will result in a 400 Bad Request response code.
```
thomasmahoney: Everything in this paragraph starting from "At present" is no longer true -- we made desired_enrichments mandatory last quarter. Could you remove this from RPAN too please?

JobParseAndNormalize.md
glpayson: Yeah I agree it kind of sucks, but we call it "4.0" in the response because we decided we didn't want to call it "skill": "careerbuilder". 

```
      "skills": {
        "textkernel": [
          {
            "name": string,
            "skill_code": integer,
            "required": boolean,
            "skill_group": string
          },
          ...
        ],
        "4.0": [
          {
            "skilldid": string,
            "normalized_term": string,
            "confidence": string,
            "type": string
          },
          ...
        ]
      },
```

Basically, I need to call it something besides ```skills``` then. Thoughts?

JobParseAndNormalize.md
glpayson: Hmm yeah I copied lots of this from the PAN documentation as you can probably tell. It the same wrong link there so I'll go ahead and change it.

JobParseAndNormalize.md
glpayson: Changing in PAN as well.

JobParseAndNormalize.md
glpayson: changed

JobParseAndNormalize.md
glpayson: I liked your suggestion better there anyway so I went with it.

JobParseAndNormalize.md
glpayson: more copy pasta. Changed in PAN as well.

JobParseAndNormalize.md
thomasmahoney: hmm... I see, yeah that is a problem. How about this?
```
- Skills Extraction
  - Textkernel Skills Extraction: The job parser will extract normalized skills using Textkernel's extraction engine and skills taxonomy by default. These skills are returned as part of the base parsing response (regardless of desired_enrichments) and are unversioned.
  - CareerBuilder Skills Extraction: When "skills" is requested as a desired_enrichment, the service will enrich the parsed job data using the CareerBuilder Skills API [insert github link here].

File: JobParseAndNormalize.md (hunk 2 above)
Comment (thomasmahoney) [reply JobParseAndNormalize.md]: man, we're the greatest at maintaining our documentation. thanks for fixing this + other boo-boos

File: JobParseAndNormalize.md [hunk 8]
```diff
@@ -41,11 +40,11 @@ This service supports the HTTP GET and POST methods. Because Base64-encoded docu
 
 The following parameters may be supplied in the query string (for HTTP GET) or form body (for HTTP POST):
 
-* **document** (Required) -- A .doc, .docx, .pdf, .rtf, .txt, .odt, .wps, and .pages documents given in a BASE64 encoded string.  Please note that the .pages format is not accepted by Textkernel; you will need to specify another parser.
+* **document** (Required) -- A .doc, .docx, .pdf, .rtf, .txt, .odt, .wps, and .pages document given in a BASE64 encoded string.  Please note that the .pages format is not accepted by Textkernel; you will need to specify another parser.
```
thomasmahoney: still need to remove refs to .pages (we only supported those in rpan for daxtra/HA)

ParseAndNormalize.md
```diff
@@ -12,7 +12,7 @@ _____________
...
+This service is an enhanced version of our [Resume Parsing API](https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/ResumeParsing.md) that parses a Base64-encoded resume and further enriches the parsed data by issuing calls to several other classification APIs. Specifically, the service currently makes one call to our [Geocoding API](https://github.com/careerbuilder/DataScienceAPIDocumentation/blob/master/Geocoding.md) to normalize the candidate's location information; one call to the [Skills API](https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/Skills.md) to parse skills from the complete text of the resume; and one call per work history item to the [Job Level](https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/JobLevel.md), [Job Title](https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/JobTitle.md), [Company Normalization] (https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/CompanyNormalization.md), and [Skills](https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/Skills.md) APIs to enrich each work history accordingly.
```
thomasmahoney: > This service is an enhanced version of our Resume Parsing API

thomasmahoney: (also, rampant bad copypastas regarding resumes, work histories, etc here)

ParseAndNormalize.md
```diff
@@ -12,7 +12,7 @@ _____________
 Summary
```
thomasmahoney, line 12: we should probably rename this file to ResumeParseAndNormalize.md too

glpayson: done

JobParseAndNormalize.md
glpayson: done. Also changes 'and' to 'or' here.

JobParseAndNormalize.md
```diff
@@ -13,21 +13,21 @@ _______
...
- - Employment Level
- - Contract Type
+ ####Available Enrichments:
+ - __Job Title Classifications__
+    - __[Carotene](https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/JobTitle.md)__ <sup>(optional)</sup>
+    - __[ONET](https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/JobTitle.md)__ <sup>(optional)</sup>
+    - __Textkernel__
+ - __[Normalized Company](https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/CompanyNormalization.md)__  <sup>(optional)</sup> - Normalized company data including name, website, etc.
+ - __Skills Extraction__
+    - __Textkernel Skills Extraction__ - Normalized skills using extracted using Textkernel's extraction engine and skills taxonomy. These skills are returned by default as part of the response (regardless of desired_enrichments) and are unversioned.
```
thomasmahoney: "Normalized skills using extracted using"

thomasmahoney (approved):

gliu6 (commented):
👍

PR #66, opened 2017-01-24
add relevant skills

andhesmi:


Skills.md
```diff
@@ -38,6 +39,24 @@ JSON
   ]
 }
 ```
+If the return_relevant_skills request parameter is set to true, the reponse will look like this:
+
+```
+{
+	"response": [{
```
thomasmahoney: we should fix up the formatting for readability, and I would do something like this to indicate that the relevant_skills list is longer than shown:
```
"relevant_skills": [
  {
  }, (...)
]

File: Skills.md [hunk 2]
```diff
@@ -13,6 +13,7 @@ This API currently supports versions 4.1 and 4.2. Version is passed in the Accep
...
 | threshold      | No | A double value between 0 and 1 controlling minimum relevancy scores for skill recognition. Higher values will more tightly restrict the returned skill tags. Default is 0.5. A threshold of 0 means all seed skill phrases recognized by exact string matching will be returned. Note that this parameter is only supported for inputs in English. |
 | auto_thres     | No | A boolean value indicating whether automatic thresholding is desired. If enabled, then when input text contains 150 or fewer words, the threshold parameter will be overwritten to 0 and all confidence values will be overwritten to 1.0&#42;. Default is true.  Note that this parameter is only supported for inputs in en,fr, and de. |
+| return_relevant_skills | No | A boolean value indicating whether to return relevant skills for each extracted skill. Defaults to false. Relevant skills are identical to extracted skills, with the confidence rating their relevance to the extracted skill. |
```
thomasmahoney, line 16: What did we decide on for a max # of relevant skills per skill? We should document that too.

andhesmi: I added all the relevant skills that were normalized skills, with their confidences. So no limit at the moment. I can change that to 10 if you'd like.

Skills.md
andhesmi: hmm.. this is the jsonlint format. Is there another json validator you use for formatting?

gliu6: I would recommend this one:
http://www.freeformatter.com/json-formatter.html#ad-output
I love this since it gives you different format options. Especially, I like the one that outputs to one compact line, so you can easily put in unit tests.

Skills.md
gliu6: I think the # of relevant skills may vary differently on skills, some may simply have more siblings while others don't. Instead of using a max # of relevant skills, I would like to suggest a base threshold (all relevant skills higher than that will be returned), but the problem is how to determine that, whether we fix it or ask user to supply parameters.

andhesmi: I thought about that Gavin, but I looked at the results and turns out the confidences tend to vary in tight groups (probably an artifact of the scoring?). So for example some skills have relevant skills all between 0.9 and 0.8, others all between 0.7 and 0.6. If we added a threshold, we'd get many skills where the threshold has no effect, and many skills with no relevant skills. If we need to threshold I think it makes more sense to take the top N.

thomasmahoney:
👍

gliu6:
👍

PR #75, opened 2017-02-07
removing job location city

glpayson:
Company location and job location share a common response object. In practice though, job location city will never be returned because their is no city on the corresponding tk response fields.


andhesmi:
i think we should have a meeting about this

andhesmi:
:+1:

thomasmahoney (approved):

PR #76, opened 2017-02-10
remove DDC references

thomasmahoney:


matthuhman:
👍

gliu6 (commented):
👍 
Just one thought, will it be better to include the IER team's contact.

PR #78, opened 2017-02-15
initial Overrides documentation

andhesmi:


SemanticOverrides.md
```diff
@@ -0,0 +1,85 @@
...
+
+User overrides are added and deleted using a RESTful interface.
+
+###HTTP method PUT
+
+Adds or updates an override for the specified user_id, term, language, taxonomy, and enrichment combination. 
+Since this endpoint modifies override state, it is restricted to certain OAuth client ids. For questions about access, send an email to 
+DataScienceApplicationDevelopment@careerbuilder.com
+
+Parameters (URL-encoded path parameters):
```
thomasmahoney: this API relies heavily on usage of path parameters, which is slightly more complex than the usual URI query params / form urlencoded params most of our APIs use. I think it might be nice to demonstrate the structure here (without example values, just URI template-style). something like [...]/user/{user_id}/{term}/[...]

SemanticOverrides.md
```diff
@@ -0,0 +1,85 @@
...
+|RawJobTitlesJobSeekerV1 | Name |
+|RawJobTitlesRecruiterV1 | Name |
+|CustomKeywords| Name |
+|ONet17| Id |
+|CaroteneV3 | Id |
+|SkillsV4 | Id |
+
+Parameters (Body parameters):
+-         returned (required): boolean, defaults to false. Whether to return this enrichment, under this taxonomy, in queries containing this term and for this language and user_id.
+-         selected (required): boolean, defaults to null. Sets state of the "selected" flag in the Semantic Search response for this enrichment, under this taxonomy, in queries containing this term and for this language and user_id.
```
thomasmahoney: a boolean defaults to null?

andhesmi: yes. its technically a Boolean

thomasmahoney (approved):

SemanticOverrides.md
```diff
@@ -0,0 +1,89 @@
...
+Parameters:
+
+Identical to PUT parameters, except that body parameters are ignored. 
+
+Example Request
+```
+HTTP DELETE
+https://api.careerbuilder.com/core/SemanticOverrides/user/U123456789/java+developer/en_us/CaroteneV3/15.2/
+```
+
```
gliu6: Do you think it is better to include the same sentence here in line 64 to make it clear?

SemanticOverrides.md
```diff
@@ -0,0 +1,89 @@
...
+Example Request
+```
+HTTP PUT
+https://api.careerbuilder.com/core/SemanticOverrides/user/U123456789/java+developer/en_us/CaroteneV3/15.2/
+Body:
+returned=true&selected=true
+```
+
+The response will include no data unless an error occurred, in which case the response will be in CB API standard format.
```
gliu6, line 64: I wonder is there an link page to official CB API standard format. I know our standard, but not CB standard.

gliu6 (commented):
very well written.

SemanticOverrides.md
andhesmi: it's somewhere, I don't think I need to link to it here though.

thomasmahoney: yeah, I just have it saved on my computer as a PDF. If platformsoftware ever makes it linkable (puts it on github or swagger or somewhere else), we could definitely link to it in a "General"/"Getting Started" page on this repo. but that's a task for another day

gliu6:
👍

SemanticOverrides.md
```diff
@@ -0,0 +1,90 @@
...
+https://api.careerbuilder.com/core/SemanticOverrides/user/U123456789/java+developer/en_us/CaroteneV3/15.2/
+Body:
+returned=true&selected=true
+```
+
+The response will include no data unless an error occurred, in which case the response will be in CB API standard format. 
+
+###HTTP Method DELETE
+
+Removes and override for the specified user_id, term, language, taxonomy, and enrichment combination. The Semantic Search API will return its default behavior for this user_id/language before this override was added.
```
msundlof, line 68: "an override" not "and override"

PR #79, opened 2017-02-17
Virus scanner doc added

matthuhman:


VirusScanner.md
```diff
@@ -0,0 +1,53 @@
+#VirusScanner API
+
+&nbsp;
+
+#Contents
+
+- [Description](#description)
+- [Request](#request)
+- [Request Structure](#geocoding-request-structure)
```
thomasmahoney: bad copy/paste from geocoding here

VirusScanner.md
```diff
@@ -0,0 +1,53 @@
...
+- [Response Structure](#response-structure)
+- [Versioning](#versioning)
+
+----------------------------
+#Description
+
+This service runs advanced antivirus scans on any file type. The virus definitions that the service is based on are updated every five minutes to ensure that the service is as up to date as possible.
```
thomasmahoney: "updated every five minutes from Avira's virus definitions repository" might give a little more information here

VirusScanner.md
```diff
@@ -0,0 +1,53 @@
...
+
+The libraries that these virus scans are based on are being constantly updated by Avira. We recommend regular reflows of data to ensure that your data security is as up to date as possible. All files that return **is_safe = false** should be immediately quarantined and the source of the data should be investigated.
```
thomasmahoney: "are being" --> just "are"? and maybe "consistently" or "regularly", "constantly" sounds slightly misleading haha. Good bit about the reflows though.

VirusScanner.md
```diff
@@ -0,0 +1,53 @@
...
+
+----------------------------
+#Request
+
+A Virus Scanner request is simply the base64 string representation of a file to be scanned.
+
+#Request Structure
+
+| Parameter  | Description |
```
thomasmahoney: you maybe should also add that this service supports GETs and both form-encoded and json POSTs (but the latter is recommended due to size limitations on GETs)

VirusScanner.md
```diff
@@ -0,0 +1,53 @@
...
+
+----------------------
+#Response Structure
+
+All responses with an HTTP status of 200 will consist of a JSON object with a top-level "data" node containing the following elements:
+
+| Field    | Description |
+|----------|-------------|
+| is_safe  | A **boolean** value to quickly tell you whether or not the document is believed to contain a virus. |
+| virus_definition | A **string** value that will always be returned. If no virus is found, this will say "200 OK" - if a virus is found, this field will contain more specific information about the nature of the virus. |
```
thomasmahoney, line 41: Not a request for any change here but just curious, will this always be either "200 OK" or "310 Virus Found"?

matthuhman: I can't remember exactly what it looks like off the top of my head, but the message will actually contain specific information about the type of virus found if that information is returned by Avira. For example, it could say something about it being a Word macro.

PR #92, opened 2017-05-17
(WIP) Semantic Search v2.0 updates (from Ruben)

carschno:
(cherry picked from commit 73309fa)

### DIFF
--- a/SemanticSearch.md
+++ b/SemanticSearch.md
@@ -1,7 +1,9 @@
 Semantic Search API
 ===================
 
-The semantic search API supports two endpoints: /query and /document. The /query endpoint interprets and provides related entities for queries. The /document endpoint uses a document's title and content to extract relevant keywords and find related entities.
+The semantic search API supports two endpoints: /query and /document.
+The /query endpoint interprets and provides related entities for queries.
+The /document endpoint parses a document, either a job or resume, and returns the most relevant semantic entities.
 
 # Table of Contents
 _____________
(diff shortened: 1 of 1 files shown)


gliu6:
Is there any update commit, I put a bunch of comments, and they are all gone.

gliu6:
Discussed with our team, as you know, we cannot remove the documentation for the existing 1.0, which is still used by people, and compared with 1.0 /document endpoint, 2.0 behaves very differently. So we could put the 2.0 documentation as a separate file. An we provide a summary readme within which has two links, one to the old 1.0 and another to 2.0, which is this file.

SemanticSearch.md
```diff
@@ -141,91 +140,158 @@ The query response is divided into two parts. First is the parsed_input node, wh
...
+		],
+		"versions" : {
+			"job_titles" : "CaroteneV3",
+			"text_kernel_related_keywords" : "TextKernelRelatedSearchTerms",
+			"job_level" : "JobLevel",
+			"occupations" : "ONet17",
+			"related_keywords" : "RelatedSearchTermsV1",
+			"extracted_keywords" : "ExtractedKeywords",
+			"skills" : "SkillsV4",
+			"custom_keywords" : "CustomKeywords"
```
gliu6, line 294: Note that this response is already outdated, we don't have joblevel anymore.
Besides, we need to do a sumary user know that the extracted_keywords in the response has two kinds of entities, one kind is from solr(name, type, weight, relationship), another kind is from parsed data (from PAN) that has (type and entity_node). currently we have two PAN entities (Location and experience level).

SemanticSearch.md
```diff
@@ -236,3 +302,54 @@ The document response is divided into two parts. First is the extracted keywords
...
+	- Removed the 'title' request parameter.
+	- Removed the 'content' request parameter.
+	- Added the 'document' request parameter (base64 encoded binary file).
+	- Added the 'type' request parameter ("JOB" or "RESUME").
+	- Removed the 'summary' field from the response.
+	- Removed the sub field 'id' from the 'extracted_keywords' response field.
+	- Added the 'parsed_input' field to the response.
+	- Added relationships to the extracted_keywords.
+
+## API changes from 0.8 to 1.0
```
gliu6, line 352: this could be moved since we are doing a separate doc beside 1.0

Geerlings:
Please reject and close this merge request. I am about to create a new one.

PR #93, opened 2017-05-18
Ltr docs

thomasmahoney:


LikelihoodToRespond.md
```diff
@@ -0,0 +1,70 @@
+Likelihood To Respond Service
+=============
+
+Table of Contents
```
gliu6: I would just call it Contents since there is no table.

LikelihoodToRespond.md
```diff
@@ -0,0 +1,70 @@
...
+_________
+- [Overview](#overview)
+- [Request Information](#request-information)
+- [Response Information](#response-information)
+- [Sample Request](#sample-request)
+
+# Overview
+
+The Likelihood To Respond (LTR) service provides data to describe the likelihood that a job seeker will respond positively when contacted regarding a new job opportunity. Data returned by the service is computed ahead of time by the Search Data Science team and is refreshed nightly.
```
gliu6: Not sure here, but will it be "regarding to"?

LikelihoodToRespond.md
```diff
@@ -0,0 +1,70 @@
...
+
+**Generic threshold queries**<br />The service may be queried with an `ltr_threshold` decimal parameter to retrieve records for job seekers whose generic LTR score matches or exceeds the provided threshold. The `ltr_threshold` parameter must be between 0.0 and 1.0, inclusive.
+
+The `skip_results` integer parameter may optionally be provided to enable "stepping through" a large result set. If no records meet or exceed the provided threshold, an empty list will be returned.
+
+**Carotene-specific threshold queries**<br />The service may be queried with an `ltr_threshold` parameter accompanied by a `carotene_id` parameter. The `carotene_id` parameter should be a valid and complete CaroteneV3 ID. (A full list of CaroteneV3 IDs can be obtained via the [Taxonomy Service](/TaxonomyService.md).) No validation is performed on the provided Carotene ID.
+
+The `skip_results` integer parameter may optionally be provided to enable "stepping through" a large result set. If the Carotene ID does not match any records, or none of the matched records meet or exceed the provided threshold, an empty list will be returned.
```
gliu6: this looks redundant.  I wonder this `skill_results` behaves differently with and without caroteneID?

LikelihoodToRespond.md
```diff
@@ -0,0 +1,70 @@
...
+
+Invalid requests (such as those specifying both `email_address` and `ltr_threshold`, or specifying an `ltr_threshold` value that is not a decimal in the specified range) will receive an HTTP 400 Bad Request error.
```
gliu6: This looks a bit redundant, too. I think it is better to mark which is required, which is optional. And if users don't follow, they should expect 400.

LikelihoodToRespond.md
```diff
@@ -0,0 +1,70 @@
+Likelihood To Respond Service
+=============
+
+Table of Contents
+_________
+- [Overview](#overview)
+- [Request Information](#request-information)
+- [Response Information](#response-information)
+- [Sample Request](#sample-request)
```
msundlof: Where is the sample request?

LikelihoodToRespond.md
```diff
@@ -0,0 +1,70 @@
...
+* `results`: An array of LTR records. May be empty. For email lookup requests, contains a maximum of one element. For both types of threshold query requests, contains a maximum of 1000 elements. (The `skip_results` request parameter may be used to obtain results beyond the first 1000 for a query.)
+* `email_address`: The job seeker's email address.
+* `popularity`: A normalized score between 1 (least popular) and 3 (most popular) that describes the frequency with which recruiters have recently performed actions on this job seeker's profile.
+* `likelihood_to_respond`: A score between 0.0 (lowest likelihood) and 1.0 (highest likelihood) indicating how likely the job seeker is to respond positively to contacts regarding new job opportunities. This is an overall score that does not consider specific job titles or classifications.
+* `likelihood_to_respond_by_carotene`: A map from carotene IDs to LTR scores. Contains between one and three elements describing the carotene ID(s) for which the job seeker is most likely to respond.
+* `activity_level`: A normalized score between 1 (least active) and 5 (most active) describing the job seeker's recent activity level on CareerBuilder sites.
```
gliu6: how about [1, 5]?

msundlof:
think it might be helpful to make a full list of possible query parameters somewhere? As of this moment I have to read through the documentation to find them. As a user of the API I'd probably like a quick list to glance at while I'm building my request.

LikelihoodToRespond.md
thomasmahoney: oh, I meant to take this line out for now. I think what I'll do instead is just put an example query string for each result type in the Request Information section

LikelihoodToRespond.md
thomasmahoney: "when contacted regarding to a new job opportunity?" nah that wouldn't be correct. maybe you're thinking of "in regards to"?

LikelihoodToRespond.md
thomasmahoney: It is redundant, in the sense that I have the same info under both request types. I feel like that's a pretty common thing to do, though. I deliberately wanted each request description to be fully fleshed out on its own -- you shouldn't have to scan sections you don't care about to get 100% up to speed on the parts you do care about.

LikelihoodToRespond.md
thomasmahoney: same as above. and it's not so simple as "this param is required and this is optional" or I would've expressed it that way. the three parameters you use for structuring your requests are email_address, ltr_threshold, and carotene_id. You can send them in any of the following combinations, but no others:
1) email_address only
2) ltr_threshold only
3) ltr_threshold and carotene_id

LikelihoodToRespond.md
thomasmahoney: hmm, I'm not so sure that's better for reader-friendly documentation? I kind of like using the parentheticals to express what the lower/higher values mean, too. Using brackets to describe a closed range is a bit more compact but feels somewhat out of place in documentation IMO.

LikelihoodToRespond.md
thomasmahoney: That's true, the "Table" part is a lie.

thomasmahoney:
@msundlof just updated with example queries. now that I think about it more (and now that I have those in place), I'm having second thoughts about listing the params right off the bat. I'm worried that readers will jump to scanning that list first, and then they'll make up their own ideas about which combinations of parameters are/aren't viable, and then probably run into trouble or confusion later. I'm thinking it might be better for the request parameter information to be embedded in the descriptions of the request types, as that will force readers to absorb knowledge of request types, not just blindly grab the list of request params and alt-f4 out. thoughts?

LikelihoodToRespond.md
gliu6: yeah, I was confused about when to use "to", now I got it, when regard plays as a verb, no to needed.

LikelihoodToRespond.md
gliu6: Would it be better to put "skip_resutls" into the above paragraphs since anyway they are not standalone parameters and you always want users to look at it. I mean like add them in () and append them to the paragraphs that need them?

LikelihoodToRespond.md
gliu6: If the above three are the only valid combinations, is it better to show these to the users instead of showing the `Invalid requests` to the users?

LikelihoodToRespond.md
gliu6: Aye.

msundlof:
👍

gliu6:
👍

PR #94, opened 2017-05-31
Add SemanticSearch API V2 documentation

Geerlings:
Now V1 and V2 have different pages.
There is one summary page that contains an overview of versions and upgrade information.

### DIFF
--- a/SemanticSearch2.md
+++ b/SemanticSearch2.md
@@ -1,0 +1,320 @@
+Semantic Search API V2
+=====================
+
+The semantic search API supports two endpoints: /query and /document.
+The /query endpoint interprets and provides related entities for queries.
+The /document endpoint parses a document, either a job or resume, and returns the most relevant semantic entities.
+
+# Table of Contents
+_____________
+
+## Query Endpoint
+
+- [Query Description](#query-description)
+
+- [Query Request Information](#query-request-information)
+
+- [Query Response](#query-response)
+
+## Document Endpoint
+
+- [Document Description](#document-description)
+
+- [Document Request Information](#document-request-information)
+
+- [Document Response](#document-response)
+
+## General Information
+
+- [Error Handling](#error-handling)
+
+# Query Description
+
+Interprets the meaning of a user's query. Used to parse a query into the intended phrases, to identify the type of entities being searched for (job title, skill, company, location, etc.), and to identify the most related other phrases/entities that help better represent the intended query. This service also applies stored user overrides to queries when user parameters are supplied in conjunction with the Semantic Overrides API. 
+
+
+# Query Request Information
+
+
+HTTP method: GET or POST
+Parameters (query/form):
+-	query (required) : query to parse and from which to infer meaning.
+-	language (optional) : two letter language code followed by underscore, followed by two letter country code. Determines the language in which enrichments are returned. Required to use personalized overrides. Defaults to en_us. Currently, allowed values are en_us, en_gb, fr_fr, de_de, and nl_nl.
+-	user_id (optional) : user id for which to look up personalized overrides. Defaults to null. 
+-	relationships_threshold : the mimimum weight of a relationship entity to be added as enrichments. Can be used to prune the result size. Defaults to 0.5. Allowed values are any number between 0 and 1.
+ 
+Example 2.0 request: 
+```
+HTTP GET
+Accept: application/json;version=2.0
+https://api.careerbuilder.com/core/semanticsearch/query?query=registered%20nurse&language=en_us&user_id=U1234
+```
+
+# Query Response
+
+The query response is divided into two parts. First is the parsed_input node, which gives information about parsing of extracted keywords in the query (this node is missing in the document response).
+Second is the extracted_keywords node, which gives the type and related entities of each extracted keyword. Enrichments can be overridden to be returned with a selected=true flag for a given user_id through the Overrides API.
+
+2.0 response:
+```
+ {
+	"data": {
+		"parsed_input": {
+			"input": "keyword",
+			"parsed": "{extracted keyword}",
+			"raw_parsed_input": "{keyword}",
+			"input_to_extracted_keywords": {
+				"keyword": "extracted keyword"
+			},
+			"parsed_fragments": ["extracted keyword"]
+		},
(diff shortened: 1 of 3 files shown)


SemanticSearch2.md
```diff
@@ -0,0 +1,320 @@
...
+			"job_titles": "CaroteneV3",
+			"skills": "SkillsV4",
+			"related_keywords": "RelatedSearchTermsV1",
+			"text_kernel_related_keywords": "TextKernelRelatedSearchTerms",
+			"custom_keywords": "CustomKeywords"
+		}
+	}
+}
+```
+# Error Handling
```
carschno, line 294: @gliu6 could you check if the error messages are still in accordance to your changes on that?

carschno:
It seems like the JSON API of the TK vacancy/CV parser should also somehow go into the documentation so that users can send a JSON doc to the `/document` endpoint.

Gavin: agreed. @carschno could you work with Ruben on adding this info in the document?

After a second thought: since the JSON API seems to be too complex to be exposed to actual clients, it should remain for internal testing purposes only. I think we should decide on how to proceed with Hai's use case before updating the documentation accordingly.

SemanticSearch2.md
```diff
@@ -0,0 +1,320 @@
...
+
+On invalid input the service returns an error message with HTTP status code 400.
```
gliu6: should add a comma between service and returns,

SemanticSearch2.md
gliu6: I think we could remove this section. All the invalid input will receive Bad Request 400. Missing parameters and empty requests are bad requests, too actually. Actually it may be a better idea to remove the documentation for the error messages since we may update that and forgot to update accordingly in the documentation.

What we need to ensure is to provide detailed information about how to use the parameters, I saw you already did. users who don't supply the required parameters in the correct way should expect an bad request.

gliu6:
I recently made a change in the old semanticSearch document, so there is a conflict, you might want to resolve that.

SemanticSearch2.md
```diff
@@ -0,0 +1,320 @@
...
+Accept: application/json;version=2.0
+https://api.careerbuilder.com/core/semanticsearch/query?query=registered%20nurse&language=en_us&user_id=U1234
+```
+
+# Query Response
+
+The query response is divided into two parts. First is the parsed_input node, which gives information about parsing of extracted keywords in the query (this node is missing in the document response).
+Second is the extracted_keywords node, which gives the type and related entities of each extracted keyword. Enrichments can be overridden to be returned with a selected=true flag for a given user_id through the Overrides API.
+
+2.0 response:
```
gliu6: We need to add additional information besides a sample response. I left a comment in Carsten's pr of documentation. Ill repost it here. We need to explain more about the fields that appeared in the response. There are two kinds of extracted keyword entities, one kind is from solr, one kind is from PAN. we need to document them here. Ruben, you can ask @carschno for a list of PAN entity enrichment.

SemanticSearch2.md
```diff
@@ -0,0 +1,320 @@
...
+- data
+	- parsed_input
+		- input (comma-separated list of extracted keywords and phrases)
+		- parsed_fragments (input as list of strings)
+		- input_to_extracted_keywords (map of extracted strings to canonical strings)
+	- extracted_keywords (list)
+		- name (canonical string)
+		- weight (float)
+		- type (string: skill, job_title, keyword, company, school, location, company_geography)
+		- entity_node (object of type-specific attributes for type: location and company_geography)
```
gliu6, line 179: I would prefer a little more information about these fields, entity_node only appears sometimes in a PAN entity, solr entity does not have this. it could be better if we could explicitly distinguish these two kinds of extracted keywords. 
Like if it is solr entity, it has name, weight, type, and relationships, for PAN entity, it has name, type, and possibly entity_node, you can ask @carschno for a list of PAN entities that we currently support.

SemanticSearch2.md
Geerlings: Currently there are 3 error types in use: "Bad request", "Missing parameters", and "EmptyRequest".

There is 1 error code in use: 0.

I don't know if there is a company-wide policy on the use of error types and error codes that would influence the design but from what you propose we could reduce the number of error types to one: "Bad request" and put all variations of missing, empty and wrong parameters under it.

SemanticSearch2.md
Geerlings: I think you mean the comma should go between 'input' and 'the'?

SemanticSearch2.md
Geerlings: That is a good point, and the same goes for the list of relationships per type.

Is it possible to define all entity_node attributes and relationships per type now? Will this no longer change?

SemanticSearch2.md
gliu6: Yeah, we only throw 400 BadReqeust as long as that is because of inappropriate user input including unsupported parameter values, missing parameters, etc. You can surely just document BadRequest briefly.

gliu6:
I saw you merge my code change, but weird, there is still a conflict. Did you pull the changes from master and then resolve the conflicts, or you just copy the changes? If you do the first way, the conflict files should be gone. Or you can click the "Resolve conflicts" button to solve the conflicts.

SemanticSearch2.md
gliu6: I think right now, it is pretty stable. You can ask Carsten for the recent changes and what Pan Entity types we are supporting now.

thomasmahoney:
@Geerlings this PR has been stagnating for a while now -- are there still outstanding work items for SS API 2.0 that prevent us from wrapping up this work? if so, then it's not ready to be pulled in to master yet, so we should close the pull request (but leave the branch intact so we can continue adding to it)

Geerlings:
Merge conflicts are fixed.

I would say: let's merge this request now.
The only remaining peer review point is to add more detail on the entity types and relationships, which is something we can do in a separate merge request (and better be done by someone else, @carschno ?).

carschno (approved):
I agree with @Geerlings  we should merge this now and make detail changes incrementally, to make it easier to submit PRs.

Geerlings:
Thanks!

PR #99, opened 2017-06-19
Refine Semantic 2.0 documentation

carschno:
- Add entity_node details
- Remove Error Handling

### DIFF
--- a/SemanticSearch2.md
+++ b/SemanticSearch2.md
@@ -169,17 +169,17 @@ The document response is divided into two parts. First is the parsed_input node 
 
 - data
 	- parsed_input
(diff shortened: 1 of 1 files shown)


SemanticSearch2.md
```diff
@@ -176,7 +176,9 @@ The document response is divided into two parts. First is the parsed_input node
 		- name (canonical string)
```
gliu6: The following format may be clearer.
- filed name (type): explanation.
So this will be -name (String): canonical form.
Also apply to the below fields.

carschno: @gliu6 updated

gliu6 (approved):

PR #101, opened 2017-06-27
Rename SemanticSearch1.md to SemanticSearchV1.md

gliu6:


carschno (approved):

SemanticSearchV1.md
```diff
@@ -1,6 +1,8 @@
 Semantic Search API V1
 ======================
 
+**Notice: The semantic search API's most updated version is [v2.0](https://github.com/careerbuilder/DataScienceAPIDocumentation/blob/master/SemanticSearchV2.md). Customers are encouraged to adjust to v2.0 since v1.0 will be deprecated in the near future.**
```
glpayson: > The semantic search API's most updated version is [v2.0]

Change to: 

> The semantic search API's current version is [v2.0]

glpayson: > Customers are encouraged to adjust to v2.0 since

Maybe: 

> Customers are encouraged  begin using v2.0, since

SemanticSearchV2.md
```diff
@@ -149,7 +145,9 @@ Parameters (query/form):
 -	user_id (optional) : user id for which to look up personalized overrides. Defaults to null. 
 -	relationships_threshold : the mimimum weight for a relationship entry to be added as enrichment. Can be used to prune the result size. Defaults to 0.5. Allowed values are any number between 0 and 1.
 -	document : the binary document to parse as base64 encoded string (according to RFC 3548/4648).
--	type : the type of document either "JOB" or "RESUME".
+-	type : the type of document is either "JOB" or "RESUME".
+-	job_title (optional, only valid when type is JOB) : a pre-define job title to override the one extracted from a job document. Notice that the job title may also be cleaned internally in order to get enrichments.
```
glpayson: a pre-define[d] job title [which] override the [job title] extracted from a job document

glpayson: I would consider losing the word 'internally'

SemanticSearchV2.md
```diff
@@ -149,7 +145,9 @@ Parameters (query/form):
...
+-	location (optional, , only valid when type is JOB): a pre-defined location to override the one extracted from a job document.
```
glpayson: You have two commas after 'optional'

a pre-defined location [which] override[s] the [location] extracted from [the] job document.

glpayson (approved):

SemanticSearchV2.md
```diff
@@ -149,7 +145,9 @@ Parameters (query/form):
 -	user_id (optional) : user id for which to look up personalized overrides. Defaults to null. 
 -	relationships_threshold : the mimimum weight for a relationship entry to be added as enrichment. Can be used to prune the result size. Defaults to 0.5. Allowed values are any number between 0 and 1.
 -	document : the binary document to parse as base64 encoded string (according to RFC 3548/4648).
--	type : the type of document either "JOB" or "RESUME".
+-	type : the type of document is either "JOB" or "RESUME".
```
thomasmahoney: we should specify parameters as `(required)` when they're required. I would rephrase this one as something like:

> the type of the document. Acceptable values are "JOB" and "RESUME".

SemanticSearchV2.md
```diff
@@ -149,7 +145,9 @@ Parameters (query/form):
...
+-	job_title (optional, only valid when type is JOB) : a pre-defined job title which overrides the job title extracted from a job document. Notice that the job title may also be cleaned in order to get enrichments.
```
thomasmahoney: "Note that..." makes more sense than "Notice that..." in this context. And is "may" the right word here? (What are the conditions under which the title will / won't be cleaned?) I think it's more correct to say "will be cleaned before being used for enrichments." You may also want to include a sentence explaining which field gets overridden in the response when this value is provided. (Same for `location` on the next line)

PR #103, opened 2017-06-27
Update JobParseAndNormalize.md

gliu6:


carschno (approved):

JobParseAndNormalize.md
```diff
@@ -44,8 +44,14 @@ The following parameters may be supplied in the query string (for HTTP GET) or f
 
 * **url** (Optional) -- When parsing html, the Job Parser has custom parsing rules for [certain domains](#domains-with-custom-parsing-rules) that can improve the quality of the parse. This field allows users who are sending html to specify the url of the document and thereby improve parsing quality for domains which have custom parsing rules. Users including this field should use the full url of the document.
 
+* **job_title** (Optional) -- a job title supplied by the user, which will be used to override the job_title field in the response. The job_title_clean filed in the response will be based on this supplied value as well instead of using the parsed from the document. This is useful when user already has the job title along with the document and believes the job title is more accurate than the parsed from the document.
```
glpayson: Suggested change: 

> The job title for the document.  If supplied, overrides the ```job_title```  and ```job_title_clean``` field in the response. 

I think it goes without saying that the job title for the job will at least never be less accurate than the parsed job title. The only time I think that might not be the case is if the job title that the user has is not already 'clean' (e.g. "Truck Driver - FULL TIME - MAKE LOTS OF MONEY!!!")

JobParseAndNormalize.md
```diff
@@ -44,8 +44,14 @@ The following parameters may be supplied in the query string (for HTTP GET) or f
...
+
+* **location** (Optional) -- a location supplied by the user, which will be used to override job_location field in the response. This is useful when user already has the location along with the document and believes the location is more accurate than the parsed from the document.
+
```
glpayson: Again, I would consider losing the last sentence since I think it goes without saying.  Also you did not capitalize the first word in the first sentence.  

If we do decide to leave the final sentence then it should be:

> This is useful when [the] user already has the location[,] along with the document[,] and believes [that] the location is more accurate than the  [location] parsed from the document.

JobParseAndNormalize.md
gliu6: Thanks, Lamar. the supplied job_title may be "ruck Driver - FULL TIME - MAKE LOTS OF MONEY!!!" actually and that may still be better than the jobtitle parsed. And in this case, job_title will just be it, but job_title_clean will be cleaned.

glpayson (approved):

PR #117, opened 2017-09-28
document skill_id

thomasmahoney:


gliu6 (commented):
You will need to add `skill_id` in `related_skills` as well.

gliu6:
👍

PR #124, opened 2017-12-11
Update Skills.md

andhesmi:


Skills.md
```diff
@@ -4,22 +4,22 @@ https://api.careerbuilder.com/core/tagging/skills
 
 This API supports the HTTP/GET and HTTP/POST methods.  
 
-This API currently supports versions 4.1 and 4.2. Version is passed in the Accept header as the parameter version (Ex. Accept:application/json;version=4.1). 4.2 contains the latest taxonomy and incorporates feedback received about specific terms since the release of V4/4.1.
+This API currently supports versions 4.1, 4.2, and 5.0. Version is passed in the Accept header as the parameter version (Ex. Accept:application/json;version=4.1). 5.0 contains the latest taxonomy and incorporates feedback received about specific terms since the release of V4/4.1 and V4.2.
```
gliu6: I wound like to change the example to `Accept:application/json;version=5.0`
shouldn't it be "the feedback"?

Skills.md
```diff
@@ -4,22 +4,22 @@ https://api.careerbuilder.com/core/tagging/skills
...
 | return_related_skills | No | A boolean value indicating whether to return related skills for each extracted skill. Defaults to false. Related skills contain the same fields as extracted skills, with confidence scores indicating their relevance to the extracted skill. |
 
 &#42; A more detailed explanation of this functionality: The tagger uses “context” to define semantic relevancy.  If the input is too short (<= 150 words) to constitute a “context,” the tagger by default returns everything by direct matching, resulting in 0.0 confidence scores due to lack of context for relatedness. To avoid confusion, a pseudo score “1” is assigned to indicate “directly matched.” This feature is called “auto thresholding” (auto_thres).
 
-The V4 taxonomy can be found through the TaxonomyAPI [here](https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/TaxonomyService.md)
+The taxonomy can be found through the TaxonomyAPI [here](https://github.com/cbdr/DataScienceAPIDocumentation/blob/master/TaxonomyService.md)
```
gliu6: maybe "The taxonomies"?
I am not sure if v4.2 is added to TaxonomyService, probably not, I marked this as tech debt, just raise attention, do we consider to add v5 there after this is done?

Skills.md
gliu6: I wound like to elaborate the "feedback" a bit more. A major difference v5 has is that it returns the only exact (or partially exact match) if that exists. In such cases, user may expect less but more accurate results. I think exposure of such information in doc is useful to customers.

Skills.md
andhesmi: I imagine we should add v5 and v4.2, yes.

Skills.md
andhesmi: I think we can talk about the specific improvements in v5 in the release note. I don't think we go into detail about improvements in successive versions anywhere else. I'd rather not make this a changelog of R&D's improvements, in general it's good enough in my opinion just to say its getting better with each version.

thomasmahoney: the "ex" should just be like, "version={VERSION_NUMBER}" so we don't have to keep updating it :) and I would do it in code-font too (with the backticks) so it's clearer that we're literally showing what your code should look like.

"incorporates feedback" is grammatically correct, no "the" needed.

+1 on not making it a changelog of R&D's improvements. if they want to start publishing such a document, we can certainly link to it here.

glpayson: Historical beef comment: Is 'Ex.' a common abbreviation for example? Could we change it to 'e.g.'?

Skills.md
```diff
@@ -4,22 +4,22 @@ https://api.careerbuilder.com/core/tagging/skills
...
 | threshold      | No | A double value between 0 and 1 controlling minimum relevancy scores for skill recognition. Higher values will more tightly restrict the returned skill tags. Default is 0.5. A threshold of 0 means all seed skill phrases recognized by exact string matching will be returned. Note that this parameter is only supported for inputs in English. |
-| auto_thres     | No | A boolean value indicating whether automatic thresholding is desired. If enabled, then when input text contains 150 or fewer words, the threshold parameter will be overwritten to 0 and all confidence values will be overwritten to 1.0&#42;. Default is true.  Note that this parameter is only supported for inputs in en, fr, and de. |
+| auto_thres     | No | A boolean value indicating whether automatic thresholding is desired. If enabled, then when input text contains 150 or fewer words , the threshold parameter will be overwritten to 0 and all confidence values will be overwritten to 1.0&#42;. Default is true.  Note that this parameter is only supported for inputs in en, fr, and de. |
```
glpayson: I think a stray space got in somehow .

Skills.md
glpayson: I agree with Gavin's beef comment about pluralizing 'taxonomy'.

Skills.md
gliu6: yeah, one after `words` and there are two spaces before `Note`

gliu6 (approved):

thomasmahoney (approved):

PR #130, opened 2018-02-21
Update RPAN doc for PII scrubbing

msundlof:
Updating Rpan to reflect our policy on mismatches with PII scrubbing.

### DIFF
--- a/ResumeParseAndNormalize.md
+++ b/ResumeParseAndNormalize.md
@@ -351,7 +351,7 @@ Response Structure
 
 PII Scrubbing
 -----------
-The service makes a best attempt at scrubbing Personally Identifiable Information (i.e. Social Security Numbers and Driver's License Numbers). PII scrubbing logic is applied to the `raw_resume_text`, `clean_resume_text`, and `resume_html` fields only. Anything found to resemble a SSN or Driver's License Number in these fields will be replaced with either: "\*\*\*Social Security Number Removed\*\*\*" or "\*\*\*Driver's License Number Removed\*\*\*". Parsed data fields (such as the phone number fields) will therefore be unaffected by false-positive PII removals.
+The service makes a best attempt at scrubbing Personally Identifiable Information (i.e. Social Security Numbers and Driver's License Numbers). PII scrubbing logic is applied to the `raw_resume_text` and `resume_html` fields only. Anything found to resemble a SSN or Driver's License Number in these fields will be replaced with either: "\*\*\*Social Security Number Removed\*\*\*" or "\*\*\*Driver's License Number Removed\*\*\*". Parsed data fields (such as the phone number fields) will therefore be unaffected by false-positive PII removals.
 
 In addition, the service attempts to scrub Date Of Birth from the same set of fields. Text resembling a Date Of Birth will be replaced with "\*\*\*Date of Birth Removed\*\*\*". To see how to disable this feature please review the `return_date_of_birth` field in the [Request structure](#request-structure) section.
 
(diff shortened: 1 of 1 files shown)


gliu6 (approved):

glpayson (approved):

ResumeParseAndNormalize.md
```diff
@@ -348,6 +348,10 @@ Response Structure
 }
 ```
 
+PII Scrubbing
+-----------
+Additionally, the service makes a best attempt at scrubbing Personally Identifiable Information i.e. Social Security Numbers and Driver's License Numbers, from the raw and clean resume fields. Anything found to resemble a SSN or Driver's License Number in the resume text will be replaced with either: "***Social Security Number Removed***" or "***Driver's License Number Removed***". False positives do occur but the scrubbing occurs post parsing and the fields that may be scrubbed are still extracted and returned in their respective fields.
```
thomasmahoney: the comma in the first sentence shouldn't be there, and the "i.e. [...]" bit should probably be in parentheses.

don't we scrub resume HTML too? we should, if not. that also contains the full resume text.

don't we also scrub date of birth unless you use the optional param to return it?

I don't understand the last sentence. it sounds like you're saying "we scrub SSN and DL from the resume text fields, but we leave the SSN and DL fields intact" which 1) would be very silly and 2) is wrong because we don't actually have fields for those things.

ResumeParseAndNormalize.md
```diff
@@ -348,6 +349,10 @@ Response Structure
 }
 ```
 
+PII Scrubbing
+-----------
+The service makes a best attempt at scrubbing Personally Identifiable Information i.e. (Social Security Numbers and Driver's License Numbers). PII scrubbing logic is applied to the raw_resume_text, clean_resume_text, and resume_html fields only. Parsed data fields (such as the phone number fields) will therefore be unaffected by false-positive PII removals. Anything found to resemble a SSN or Driver's License Number in the resume text will be replaced with either: "***Social Security Number Removed***" or "***Driver's License Number Removed***". In addition the service by default scrubs information resembling a date of birth. To see how to disable this feature please review the return_date_of_birth field in the [Request structure](#request-structure).
```
thomasmahoney: The "i.e." should be part of the parenthetical. (Read the sentence if you remove the parenthetical to see if it still makes sense)

You should surround field names with `backticks` for the monospaced formatting (tells the reader that it's something pertaining to code)

Move the "Anything found..." sentence after the "PII scrubbing..." sentence and change "in the resume text" to "in these fields"

"In addition" should be followed by a comma. I'd also revise this sentence slightly: "In addition, the service attempts to scrub date of birth from the same set of fields." I'd also start a new paragraph for this part since it's kind of a different idea.

"in the [Request structure]." --> "in the [Request structure] section."

ResumeParseAndNormalize.md
```diff
@@ -348,6 +349,12 @@ Response Structure
 }
 ```
 
+PII Scrubbing
+-----------
+The service makes a best attempt at scrubbing Personally Identifiable Information (i.e. Social Security Numbers and Driver's License Numbers). PII scrubbing logic is applied to the `raw_resume_text`, `clean_resume_text`, and `resume_html` fields only. Anything found to resemble a SSN or Driver's License Number in these fields will be replaced with either: "***Social Security Number Removed***" or "***Driver's License Number Removed***". Parsed data fields (such as the phone number fields) will therefore be unaffected by false-positive PII removals.
```
thomasmahoney: sorry, one last thing I just discovered by previewing this (btw, you should always preview your changes in markdown files). you need to escape your asterisks; currently it's bold+italicizing these strings instead of showing the asterisks. it's important because people might write code that expects these exact string literals. you might want to demonstrate the DOB Removed string in the next paragraph too.

msundlof:
clean resume is not scrubbed

glpayson (approved):

PR #131, opened 2018-02-27
cnorm-id-lookups

thomasmahoney:


CompanyNormalization.md
```diff
@@ -69,13 +80,13 @@ Example: https://api.careerbuilder.com/core/normalizedcompanies?company_name=car
 }
 ```
 
-## Response Information
+## Lookup Requests
 
-The response returns a single data node which contains a list of normalized companies. These normalized companies are ordered by the confidence score. Each normalized company has a normalized company name (string), a [NAICS](http://www.census.gov/eos/www/naics/) ID (string), a NAICS description (string), a DUNS number(string), a country (string), a state (string), a city (string), a postal code (string), a website (string), a company size (int), an ID (string), and a confidence (double). Confidence scores range from 0.0 to 1.0. A single master company will be returned.
+Lookup requests specify an `id` value. This corresponds to the `id` field associated with normalized companies returned by the service. The `max_results` parameter and the two parameters controlling classifier behavior (`use_query_classifier` and `filter_by_country`) may be included, but all of these will have no effect. Other parameters from the [Search Requests](#search-requests) section may not be supplied with the `id` parameter; requests constructed in this way will result in an HTTP 400 Bad Request error.
```
gliu6: if `use_query_classifier` and `filter_by_country` does not have any effect, why worth mentioning these. I think we could remove this.

PR #133, opened 2018-03-26
Add companyaddress documentation.

glpayson:


CompanyAddress.md
```diff
@@ -0,0 +1,130 @@
...
+
+- [Summary](#summary)
+- [Request structure](#request-structure)
+- [Response structure](#response-structure)
+- [Sample response](#sample-response)
+- [Versioning](#versioning)
+
+## Summary
+
+Company Address Service (CAS) takes a request containing a company name, latitude and longitude and returns a geocoded response.
```
thomasmahoney: Is it useful to introduce the "CAS" acronym here? I would also start this sentence with the word "the".

thomasmahoney: And maybe there's better phrasing than "a geocoded response"... maybe like, "a response describing a normalized geographic location"?

CompanyAddress.md
```diff
@@ -0,0 +1,130 @@
...
+&nbsp;
+
+Each element of the returned **results** array will be formatted as follows:
+
+|   |   |
+|---|---|
+| admin_areas | A JSON array of up to 5 admin_area objects. See below for details on the structure of these objects. **This property is always present, but may be empty** if no admin areas were found. |
+| latitude | A double value specifying the location's latitude (in degrees) within the range [-90, +90]. **This property is always present.** |
+| longitude | A double value specifying the location's longitude (in degrees) within the range [-180, +180]. **This property is always present.** |
+| country | The country or region name of an address. **This property may be absent** if the location was not associated with a country. Some locations for which this occurs are "Europe", "Pacific Ocean", and "Gaza Strip" (a contested territory). |
```
thomasmahoney: should this note still be present in this service's documentation? I mean, is there any conceivable company address request that would yield a result in the Gaza Strip?

CompanyAddress.md
```diff
@@ -0,0 +1,130 @@
...
+| locality | The populated place for the address. This typically refers to a city, but may refer to a suburb or a neighborhood in certain countries. **This property may be absent**. |
+| postal_code | The post code, postal code, or ZIP code of an address. **This property may be absent**. |
+| country_code | The two-letter ISO country code. **This property may be absent**. See "Country" for examples of when this might occur. |
```
thomasmahoney: ditto the "country" question

CompanyAddress.md
```diff
@@ -0,0 +1,130 @@
...
+| formatted_address| A string containing the human-readable address of this location. Often this address is equivalent to the "postal address," which sometimes differs from country to country. (Note that some countries, such as the United Kingdom, do not allow distribution of true postal addresses due to licensing restrictions.) **This property may be absent**. |
+| location_type | The classification of the geographic entity returned. Possible values (in descending order of specificity) are STREET_ADDRESS, SUBLOCALITY, POSTAL_CODE, LOCALITY, ADMIN_AREA_2, ADMIN_AREA_1, STREET, COUNTRY, and UNKNOWN. The returned location type will reflect the most specific field present, based on the previously described order. If none of these fields are present (such as for the examples for which the Country field is absent), then the returned location type will be Unknown. **This property is always present and non-empty.** |
```
thomasmahoney: this service shouldn't ever return the larger location_types, right? like, is it the case that this service should actually always return a STREET_ADDRESS location? if so, perhaps we don't need this field, or perhaps we should add that information to the docs...

CompanyAddress.md
```diff
@@ -0,0 +1,130 @@
...
+| postcode_localities | A string array denoting all the localities contained in a postal code. **This property is only present when the result is a postal code that contains multiple localities,** and will never be empty (nor will it ever contain only one element). |
+| street_address | The official street line of an address relative to the area, as specified by the Locality, or PostalCode, properties. **This property may be absent**. |
```
thomasmahoney: should be "as specified by the `locality` or `postal_code` properties." (could you change this in geocoding too?)

CompanyAddress.md
```diff
@@ -0,0 +1,130 @@
...
+| sublocality | The neighborhood, borough, township, etc. for the address. Sublocalities are typically more specific than cities and may be returned even if the locality field is empty. **This property may be absent**. |
+| landmark | The full name of the landmark (such as a military base or island) returned by the service. **This property may be absent**. |
+| place_id | Place Id relating to the geolocation requested.
```
thomasmahoney: wow, this wording really bothers me. "whoever wrote this"--. can we change it to like, "Google's unique identifier string corresponding to the returned location."? (and also in geocoding pretty please)

CompanyAddress.md
```diff
@@ -0,0 +1,130 @@
...
+| partial_match | A Boolean value that, when true, indicates that the geocoder did not return an exact match for the original request, though it was able to match part of the requested address. You may wish to examine the original request for misspellings and/or an incomplete address.<br>Partial matches most often occur for street addresses that do not exist within the locality you pass in the request. Partial matches may also be returned when a request matches two or more locations in the same locality. For example, "21 Henr St, Bristol, UK" will return a partial match for both Henry Street and Henrietta Street. Note that if a request includes a misspelled address component, the geocoding service may suggest an alternative address. Suggestions triggered in this way will also be marked as a partial match. **This property is always present.** |
```
thomasmahoney: We're doing place_id lookups to Geocoding, right? So this field should never be `true` and we should just remove it from this service's response.

PR #136, opened 2018-05-10
Candidate reranker

msundlof:


CandidateReranking.md
```diff
@@ -0,0 +1,92 @@
...
+        - [Query](#query)
+        - [Keyword](#keyword)
+        - [Parsed Input](#parsed-input)
+        - [Filter](#filter)
+    - [Profiles](#profiles)
+- [Response structure](#response-structure)
+- [Versioning](#versioning)
+
+### Summary
+The Candidate Reranking Service is an HTTP Rest service which performs machine learned ranking on a list of candidate profiles. The service makes use of a versioned reranking-library jar maintained by Texkterkernel which applies a machine learning model trained on Talent Discovery Platform query data.
```
psequeirag: Is "Texkterkernel" spelled out correctly?

CandidateReranking.md
```diff
@@ -0,0 +1,92 @@
...
+https://api.careerbuilder.com/core/search/candidatereranker
+OAUTH credentials are **required**.
+
+A request is composed of 4 main parts:
+`reranker_config`, `source`, `query`, and `profiles`.
+
+#### Top Level
+| Param    | Type | Description
+|----------|------|--------|
+| reranker_config | String | **Required**: The `reranker_config` determines which learned model to apply to your incoming rerank request. Currently there is only one reranker model to call so while the field is required it can be left empty or filled with a String value like "RerankerV1".
```
psequeirag: It's a little odd to have a required field that can be omitted. Also, this looks more like an enum, so it's probably better to be more categorical on the accepted values for the param.

I'd recommend either mandating it to be always supplied with the `"RerankerV1"` value, or not marking it as "Required".

psequeirag: also reword 
> a String value like "RerankerV1".

to 

> the String value "RerankerV1".

CandidateReranking.md
```diff
@@ -0,0 +1,92 @@
...
+| source | String | **Required**: Excepted Values: **EDGE** or **MY_SUPPLY**. The source identifies where profile data was attained. A model can be learned based on the source providing more accurate results per each data format.
```
psequeirag: Typo: Excepted.

Hmm, given that this is an enum like the first one is, I'm thinking you could structure the doc of the first param to be like this one.

CandidateReranking.md
```diff
@@ -0,0 +1,92 @@
...
+| query | [Query Object](#query-object-parts) | **Required**: The original fully enriched query sent to SOLR for the list of Candidate Profiles.
+| profiles | List of [Profiles](#profiles) | **Required**: List of profiles to be reranked.
```
psequeirag: This param should be a Json array, right? I think it may be preferable to use the word "Array" instead of "List" in the type column to avoid casting any doubts.

CandidateReranking.md
```diff
@@ -0,0 +1,357 @@
...
+				"ranker_score" : 0.62
+			}
+		]
+	}
+}
+```
+
+### Versioning
+---------
+The current API version is 1.0. API version must be specified in the Accept header, e.g. application/json;version=1.0.
```
msundlof, line 371: I never really thought out how to return versions in the reranker. It might be a good idea to return the config as version? (ponder)

thomasmahoney: @msundlof don't mix up an application-specific data/classifier version (like, which reranker we're using) with your application-agnostic API version (like, what you accept in the Accept header and [ideally, if we were doing everything properly and to standards] return in the... Content-Type? i think? header). we did that with skills and it didn't work out well, i.e. we went from v4 -> v4.1 -> v4.2 through a mix of breaking API syntax changes and IER data/algorithm changes. we should use headers for the API version and request parameters (query/form/json) for anything specific to the service

CandidateReranking.md
```diff
@@ -32,23 +31,22 @@ OAUTH credentials are **required**.
...
-| profiles | [Profile[]](#profile-and-objec-parts) | **Required**: Array of profiles to be reranked.
+| query | [Query](#query) | **Required**: The original fully enriched query sent to SOLR for the list of Candidate Profiles.
+| profiles | [Profile[]](#profile) | **Required**: Array of profiles to be reranked.
 
-### Query and Object Parts
+### Query
 
 ----------
-Query Object parts are expected to be taken from the response of the [SemanticSearchAPI](https://github.com/careerbuilder/DataScienceAPIDocumentation/blob/master/SemanticSearchV2.md). All the following objects mirror that response.
+The Query and it's child objects are expected to be taken from the response of the [SemanticSearchAPI](https://github.com/careerbuilder/DataScienceAPIDocumentation/blob/master/SemanticSearchV2.md). All the following objects mirror that response.
```
thomasmahoney: its*
and "query" shouldn't be capitalized if it's not starting a sentence

CandidateReranking.md
```diff
@@ -32,23 +31,22 @@ OAUTH credentials are **required**.
...
 `reranker_config`, `source`, `query`, and `profiles`.
 
-#### Top Level
 | Param    | Type | Description
 |----------|------|--------|
 | reranker_config | String | **Required**: Accepted Value: **RERANKER_V1** The `reranker_config` determines which learned model to apply to your incoming rerank request. Currently there is only one reranker model to call "RERANKER_V1".
 | source | String | **Required**: Accepted Values: **EDGE** or **MY_SUPPLY**. The source identifies where profile data was attained. A model can be learned based on the source providing more accurate results per each data format.
-| query | [Query](#query-object-parts) | **Required**: The original fully enriched query sent to SOLR for the list of Candidate Profiles.
-| profiles | [Profile[]](#profile-and-objec-parts) | **Required**: Array of profiles to be reranked.
+| query | [Query](#query) | **Required**: The original fully enriched query sent to SOLR for the list of Candidate Profiles.
```
thomasmahoney: comma after "original"

"fully enriched" should *maybe* be hyphenated? (ponder)

"Candidate Profiles" should be downcased

CandidateReranking.md
```diff
@@ -74,27 +72,27 @@ Entites returned from the SemanticSearchAPI.
 
 | Param    | Type | Description
 |----------|------|--------|
-| name | String | **Required**: Name of the keyword, e.g. "Java".
-| type | String | **Required**: Type of the keyword, e.g. "skill".
-| weight | Integer | **Required**: Weight of the keyword as an int.
+| name | String | **Optional**: Name of the keyword, e.g. "Java".
+| type | String | **Optional**: Type of the keyword, e.g. "skill".
```
thomasmahoney: if this is an enumeration, should we UPPER_CAMEL_CASE it?

CandidateReranking.md
```diff
@@ -74,27 +72,27 @@ Entites returned from the SemanticSearchAPI.
...
+| weight | Integer | **Optional**: Weight of the keyword as an int.
```
thomasmahoney: what are people supposed to use here? is there an acceptable range, like [1, 100]?

CandidateReranking.md
```diff
@@ -74,27 +72,27 @@ Entites returned from the SemanticSearchAPI.
...
 | relationships | Map<String, Entities[]> | **Optional**: Map of Strings(Semantic Enrichment Type) to a list of of entities. Valid list of Map key values are: `job_titles`, `occupations`, `skills`, `job_level`, `text_kernel_related_keywords`, `interesting_keywords`, `related_keywords`, `related_keywords_recruiter`, `related_keywords_jobseeker`, `raw_job_titles_jobseeker`, `raw_job_titles_recruiter`, `extracted_kewords`, `custom_keywords`, `related_search_terms`, and `job_titles_recruiter`.
```
thomasmahoney: I don't think it looks good to use `Map<..>` (or even the word "Map" at all) when documenting a JSON API. "Map" is mostly a Java thing, so you're effectively leaking internal details/terminology from our service implementation into the documentation of its interface -- and its interface is JSON, not Java.

It might be sufficient to just use the word "dictionary" as the type. Or you may want to define a `Relationships` object further down and link to it here, like you did in some other places.

And my same comment from above about enum casing applies here (or is it just this way from SSAPI, so we can't change it here without breaking the convenience of "pass your SSAPI response directly to CRS?")

Also, typo on "extracted_kewords"

CandidateReranking.md
```diff
@@ -74,27 +72,27 @@ Entites returned from the SemanticSearchAPI.
...
 
 #### Entity
 | Param    | Type | Description
 |----------|------|--------|
-| name | String | **Required**: Name of enrichment **Entity**.
-| weight | double | **Reqired**: Weight of enriched term according to the SemanticSearchAPI.
+| name | String | **Optional**: Name of enrichment **Entity**.
```
thomasmahoney: why is Entity capitalized here? also, an example might be good (I'm not personally sure off the top of my head what an "entity" name might be)

CandidateReranking.md
```diff
@@ -74,27 +72,27 @@ Entites returned from the SemanticSearchAPI.
...
+| weight | double | **Optional**: Weight of enriched term according to the SemanticSearchAPI.
```
thomasmahoney: "according to the SemanticSearchAPI" looks odd. I think when we're referring to it in a sentence, with "the" in front of it, we should space it out: "according to the Semantic Search API"

CandidateReranking.md
```diff
@@ -103,12 +101,13 @@ The profile is built with data from either **MY_SUPPLY** or **EDGE** (identified
 #### Profile
 | Param    | Type | Description
 |----------|------|--------|
-| document_id | String | **Required**: Unique CB identifier string for the candidate.
-| recent_jobs | [RecentJobs[]](#recentjob) | **Required**: Array of **RecentJob** objects. Recent s jobs held by the candidate. Reranker requires at least one RecentJob object.
-| skills | [Skill[]](#skill) | **Required**: Array of **Skill** objects. Skills associated with the candidate.
-| years_of_experience | Double | **Required**: Total years of experience the candidate has as a double.
+| document_id | String | **Optional**: Unique CB identifier string for the candidate.
+| recent_jobs | [RecentJobs[]](#recentjob) | **Optional**: Array of **RecentJob** objects. Recent s jobs held by the candidate. Reranker requires at least one RecentJob object.
```
thomasmahoney: typo here ("Recent s jobs")

thomasmahoney: and I would start the last sentence as "**The** reranker requires [..]"

CandidateReranking.md
msundlof: @thomasmahoney I would prefer that yes but alas the service is already in use and  this is the way we set up the contract initially. If we had done that then it would have removed the need for me to create the SemanticName methods to pull the enumerated value. I think I made a lot of little mistakes like this that made my life harder...

CandidateReranking.md
msundlof: Good point. I think I'll define it as a dictionary object and break the object into another section for further detail.

CandidateReranking.md
```diff
@@ -103,12 +101,13 @@ The profile is built with data from either **MY_SUPPLY** or **EDGE** (identified
...
+| skills | [Skill[]](#skill) | **Optional**: Array of **Skill** objects. Skills associated with the candidate.
+| years_of_experience | Double | **Optional**: Total years of experience the candidate has as a double.
```
thomasmahoney: Confusing/ambiguous phrasing here (Mike, how many years of experience do you have as a double?)

thomasmahoney: `/img stunt double mike`

CandidateReranking.md
```diff
@@ -103,12 +101,13 @@ The profile is built with data from either **MY_SUPPLY** or **EDGE** (identified
...
 | normalized_education_level | String[] | **Optional**: Normalized Educations associated with the candidate i.e., Master's Degree, Bachelor's Degree etc..
 | unnormalized_education_level | String[] | **Optional**: Unnormalized Educations associated with the candidate. i.e., masters degree, Master of Science, Masters etc..
+| major | String | **Optional**: The major or field of study the candidate's degree pertains to.
```
thomasmahoney: (going full strunk & white here..) "to which the candidate's degree pertains"

CandidateReranking.md
```diff
@@ -325,7 +325,7 @@ A **Response** is composed of a array of `ranked_profiles`. Each profile has the
 | Param    | Type | Description
 |----------|------|--------|
 | document_id | String | Document Id associated with a candidate profile.
-| score | Double | Score as double value as returned from the **Reranker**.
+| score | double | Score as double value as returned from the **Reranker**.
```
thomasmahoney: "as double value" is probably redundant since we have the type info right next to it. I would just put "Score as returned from...". Also maybe include a range ([0.0, 1.0] probably?) and downcase the word "Reranker"

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+The service supports POST requests to:
+https://api.careerbuilder.com/core/search/candidatereranker
+OAUTH credentials are **required**.
+
+A request is composed of 4 main parts:
+`reranker_config`, `source`, `query`, and `profiles`.
+
+| Param    | Type | Description
+|----------|------|--------|
+| reranker_config | String | **Required**: Accepted Value: **RERANKER_V1** The `reranker_config` determines which learned model to apply to your incoming rerank request. Currently there is only one reranker model to call "RERANKER_V1".
```
glpayson: Could we have a `Required` (boolean) column, and a `Description` column?

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+
+| Param    | Type | Description
+|----------|------|--------|
+| name | String | **Optional**: Name of the keyword, e.g. "Java".
+| type | String | **Optional**: Type of the keyword, e.g. "skill".
+| weight | Integer | **Optional**: Weight of the keyword as an int. \[0-100\]
+| relationships | [Relationship](#relationship) | **Optional**: Array of relationship objects which enrich the keyword with additional entities.
+
+#### Relationship
+A `relationship` is composed of a **\<semantic enrichment type\>** and the associated set of entities which can be classified under that **\<semantic enrichment type\>**. Accepted values for the **\<semantic enrichment type\>** include: `job_titles`, `occupations`, `skills`, `job_level`, `text_kernel_related_keywords`, `interesting_keywords`, `related_keywords`, `related_keywords_recruiter`, `related_keywords_jobseeker`, `raw_job_titles_jobseeker`, `raw_job_titles_recruiter`, `extracted_kewords`, `custom_keywords`, `related_search_terms`, and `job_titles_recruiter`.
```
glpayson: Is this the full set of possible semantic enrichment types? If so then we should say "Accepted values for the **\<semantic enrichment type\>** are: ...." aot "include" which implies there are more accepted values not listed here.

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+| Param    | Type | Description
+|----------|------|--------|
+| **\<semantic enrichment type\>** | [Entity[]](#entity) | **Optional**: Relationship to the keyword `name` value.
+
+#### Entity
+An `entity` is composed of a `name`, `weight`, and `selected` boolean. An `entity` enriches the original keyword with terms or phrases which share a close association. For example, if you were to have a `keyword` java, an enrichment entity might be java or programming.
+
+| Param    | Type | Description
+|----------|------|--------|
+| name | String | **Optional**: Name of enrichment `entity`. Ex., javascript, java, programming or software engineer.
```
glpayson: Beef comment. I prefer E.g. to Ex. https://english.stackexchange.com/questions/16197/whats-the-difference-between-e-g-and-ex

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+| weight | double | **Optional**: Weight of enriched term according to the Semantic Search API.
```
glpayson: Is there a common valid range here? If so can we list it?

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+| selected | boolean | **Optional**: Specific for the `related_search_terms` Semantic Enrichment. If selected the value is sent to the Reranker as a feature to be extracted for reranking else it is left out of reranking. All entites default to true.
```
glpayson: Not really a request for change, I'm just curious why would this ever be false? This effectively means that the front end is sending us an entity we will ignore?

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+
+#### Filter
+| Param    | Type | Description
+|----------|------|--------|
+| id | String | **Optional**: Unique identfier for the **Filter**.
+| name | String | **Optional**: Name of the **Filter**.
+| type | String | **Optional**: Type of **Filter** applied. Accepted values are *facet* or *range*. Service defaults to *facet*.
```
glpayson: Beef comment, for possible values I like `this` better than *that*.

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+| values | String[] | **Optional**: Values chosen to filter on from the original search request **Filter**.
+
+### Profile
+----------
+The profile is built with data from either **MY_SUPPLY** or **EDGE** (identified in the `source` param). Using data from each profile the service builds a RerankRequest Document which is sent along with features extracted from the query to the **Reranker**. The **Reranker** then returns the list of profiles in a new ordering which should be of more relevance to the query.
+
+**IMPORTANT: Though many fields on the profile are optional its encouraged to fill out the profile as fully as possible in order to get the best results from the Reranker**.
```
glpayson: Should be "it's encouraged". Beef comment: I kind of prefer "it is" for documentation.

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+
+#### Profile
+| Param    | Type | Description
+|----------|------|--------|
+| document_id | String | **Optional**: Unique CB identifier string for the candidate.
+| recent_jobs | [RecentJobs[]](#recentjob) | **Optional**: Recent jobs held by the candidate.
+| skills | [Skill[]](#skill) | **Optional**: Array of **Skill** objects. Skills associated with the candidate.
+| years_of_experience | Double | **Optional**: Total years of experience the candidate has as a double.
```
glpayson: "as a double" is a bit redundant given the Type column.

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+| normalized_education_level | String[] | **Optional**: Normalized Educations associated with the candidate i.e., Master's Degree, Bachelor's Degree etc..
```
glpayson: Beef comment: extra period after "etc." is unnecessary.

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+| unnormalized_education_level | String[] | **Optional**: Unnormalized Educations associated with the candidate. i.e., masters degree, Master of Science, Masters etc..
```
glpayson: Same beef comment.

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+| major | String | **Optional**: The major or field of study the candidate's degree pertains to.
+| city | String | **Optional**: City that the candidate is located in.
+| state | String | **Optional**: State that the candidate lives in.
+| country | String | **Optional**: Country of the candidate.
```
glpayson: Sorry for all the beef comments :( , but could we use consistent wording on city, state and country? I would just put  "X that the candidate is located in." for all three.

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+| latitude | Double | **Optional**: Location of **Latitude** associated with the canddiates location.
+| longitude | Double | **Optional**: Location of **Longitude** associated with the candidates location.
```
glpayson: These two are a bit awkwardly worded imo. I would just put "**Latitude** of the candidate's location." Even if you don't reword it, "candidates" should be "candidate's"

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+
+#### RecentJob
+
+| Param    | Type | Description
+|----------|------|--------|
+| company_name | String | **Optional**: Company name where job was held for the candidate.
+| onet | Onet | **Optional**: ONet title associated with job.
+| carotene | Carotene | **Optional**: Carotene title associated with the job.
+| unclassified | Unclassified | **Optional**: Unclassified job title if the job does not have an ONet or Carotene title.
```
glpayson: Is this only returned if the `onet` and `carotene` are not? It sounds like that is the case, but maybe we could make that explicit?

CandidateReranking.md
msundlof: Indeed, they are sending us an entity we ignore if it is sent as false.

CandidateReranking.md
```diff
@@ -32,118 +32,118 @@ OAUTH credentials are **required**.
...
-| company_name | String | **Optional**: Company name where job was held for the candidate.
-| onet | Onet | **Optional**: ONet title associated with job.
-| carotene | Carotene | **Optional**: Carotene title associated with the job.
-| unclassified | Unclassified | **Optional**: Unclassified job title if the job does not have an ONet or Carotene title.
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| company_name | String | **FALSE** | Company name where job was held for the candidate.
+| onet | Onet | **FALSE** | ONet title associated with job.
+| carotene | Carotene | **FALSE** | Carotene title associated with the job.
+| unclassified | Unclassified | **FALSE** | If the job has no associated onet or carotene the service exceptes an unclassfied job title sent under the unclassifed object.
```
glpayson: Type here "exceptes".

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+- [Response Structure](#response-structure)
+- [Versioning](#versioning)
+
+### Summary
+The Candidate Reranking Service is an HTTP Rest service which performs machine learned ranking on a list of candidate profiles. The service makes use of a versioned reranking-library jar maintained by Textkernel which applies a model trained on Talent Discovery Platform search data.
+
+### Request Structure
+The service supports POST requests to:
+https://api.careerbuilder.com/core/search/candidatereranker
+OAUTH credentials are **required**.
```
glpayson: Are you wanting this line on a new line because if so you need another cr.

screen shot 2018-06-25 at 1 01 21 pm

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+
+A request is composed of 4 main parts:
+`reranker_config`, `source`, `query`, and `profiles`.
+
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| reranker_config | String | **TRUE** | Accepted Value: **RERANKER_V1** The `reranker_config` determines which learned model to apply to your incoming rerank request. Currently there is only one reranker model to call "RERANKER_V1".
```
glpayson: Needs a colon or comma after after "... only one reranker model to call"

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+|----------|------|----------|-------------|
+| **\<semantic enrichment type\>** | [Entity[]](#entity) | **FALSE** | Relationship to the keyword `name` value.
+
+#### Entity
+An `entity` is composed of a `name`, `weight`, and `selected` boolean. An `entity` enriches the original keyword with terms or phrases which share a close association. For example, if you were to have a `keyword` java, an enrichment entity might be java or programming.
+
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| name | String | **FALSE** | Name of enrichment `entity`. e.g., javascript, java, programming or software engineer.
+| weight | double | **FALSE** | Range: \[0.0 - 1.0\]. Weight of enriched term according to the Semantic Search API.
```
glpayson: Minor beef comment: could we put the range at the end to mirror how you specified it [here](https://github.com/careerbuilder/DataScienceAPIDocumentation/pull/136/files#diff-454d057d2679f3281a780face2d22ae4R78)?

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| id | String | **FALSE** | Unique identfier for the **Filter**.
+| name | String | **FALSE** | Name of the **Filter**.
+| type | String | **FALSE** | Type of **Filter** applied. Accepted values are `facet` or `range`. Service defaults to `facet`.
+| values | String[] | **FALSE** | Values chosen to filter on from the original search request **Filter**.
+
+### Profile
+----------
+The profile is built with data from either **MY_SUPPLY** or **EDGE** (identified in the `source` param). Using data from each profile the service builds a RerankRequest Document which is sent along with features extracted from the query to the **Reranker**. The **Reranker** then returns the list of profiles in a new ordering which should be of more relevance to the query.
```
glpayson: I think you need a couple commas in this sentence: "Using data from each profile, the service builds a RerankRequest Document, which is sent ..." 

Also need a comma on the last sentence: "The **Reranker** then returns the list of profiles in a new ordering, which should be of more relevance to the query."

CandidateReranking.md
```diff
@@ -0,0 +1,368 @@
...
+        }
+      ]
+    },
+    "filters": [
+      {
+        "id": "",
+        "name": "Freshness",
+        "type": "Range",
+        "unit_of_measurement": "days",
+        "values": [0,-1]
```
glpayson: Is this really the range? If so should it be [-1,0]?

msundlof: That's what TDP sends (shrug)

CandidateReranking.md
```diff
@@ -0,0 +1,369 @@
...
+
+OAUTH credentials are **required**.
+
+A request is composed of 4 main parts:
+`reranker_config`, `source`, `query`, and `profiles`.
+
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| reranker_config | String | **TRUE** | Accepted Value: **RERANKER_V1** The `reranker_config` determines which learned model to apply to your incoming rerank request. Currently there is only one reranker model to call, "RERANKER_V1".
+| source | String | **TRUE** | Accepted Values: **EDGE** or **MY_SUPPLY**. The source identifies where profile data was attained. A model can be learned based on the source providing more accurate results per each data format.
```
glpayson: Should be 'obtained' not 'attained'.

glpayson: Also probably need a comma after "source" in the second sentence. Perhaps better as "A model can be learned based on the source, thereby providing more accurate results per each data format."

CandidateReranking.md
```diff
@@ -0,0 +1,369 @@
...
+
+#### Relationship
+A `relationship` is composed of a **\<semantic enrichment type\>** and the associated set of entities which can be classified under that **\<semantic enrichment type\>**. Accepted values for the **\<semantic enrichment type\>** are: `job_titles`, `occupations`, `skills`, `job_level`, `text_kernel_related_keywords`, `interesting_keywords`, `related_keywords`, `related_keywords_recruiter`, `related_keywords_jobseeker`, `raw_job_titles_jobseeker`, `raw_job_titles_recruiter`, `extracted_kewords`, `custom_keywords`, `related_search_terms`, and `job_titles_recruiter`.
+
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| **\<semantic enrichment type\>** | [Entity[]](#entity) | **FALSE** | Relationship to the keyword `name` value.
+
+#### Entity
+An `entity` is composed of a `name`, `weight`, and `selected` boolean. An `entity` enriches the original keyword with terms or phrases which share a close association. For example, if you were to have a `keyword` java, an enrichment entity might be java or programming.
```
glpayson: For the example here could we maybe pick a better example enrichment entity? I feel like saying "java" is an enrichment entity for "java" keyword doesn't get us much further in explaining what an entity is.

CandidateReranking.md
```diff
@@ -0,0 +1,369 @@
...
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| id | String | **FALSE** | Unique identfier for the **Filter**.
+| name | String | **FALSE** | Name of the **Filter**.
+| type | String | **FALSE** | Type of **Filter** applied. Accepted values are `facet` or `range`. Service defaults to `facet`.
+| values | String[] | **FALSE** | Values chosen to filter on from the original search request **Filter**.
+
+### Profile
+----------
+The profile is built with data from either **MY_SUPPLY** or **EDGE** (identified in the `source` param). Using data from each profile, the service builds a RerankRequest Document, which is sent along with features extracted from the query to the **Reranker**. The **Reranker** then returns the list of profiles in a new ordering, which should be of more relevance to the query.
```
glpayson: One minor comment on "Using data from each profile". Should it be "Using data from either profile source"? Maybe I am misunderstanding, but there is only one profile source right? You can't have both a my supply and edge profile in one request?

msundlof: It would be good for me to clarify that all profiles are from the same source here or maybe not say anything about the source if that's adding confusion.

msundlof: The list of profiles comes from a single source not both.  And I got into more specific details on the service itself which I don't need to do. I think what's important to understand is that data/features are extracted from the profile to help build the new more relevant results.

CandidateReranking.md
```diff
@@ -0,0 +1,369 @@
...
+      "longitude": -84.3879824
+    }
+  ]
+}
+```
+
+### Response Structure
+----------
+
+A **Response** is composed of a array of `ranked_profiles`. Each profile has the original `document_id` and a newly aquired `ranker_score` from the **Reranker**. `ranked_profiles` will be listed in descending order according to their new scores.
```
glpayson: Should be "an array"

CandidateReranking.md
```diff
@@ -0,0 +1,369 @@
...
+
+#### RerankedProfile
+
+| Param    | Type | Description
+|----------|------|--------|
+| document_id | String | Document Id associated with a candidate profile.
```
glpayson: Should be "Document ID associated ..."

glpayson (approved):

PR #137, opened 2018-06-04
Adding README for CWA and CJT.

glpayson:


CompanyWorkActivities.md
```diff
@@ -0,0 +1,54 @@
...
+- [Response structure](#response-structure)
+- [Versioning](#versioning)
+
+## Summary
+
+The Company Work Activities service provides a list of work activities associated with a 
+`company_id` and `carotene_id` pair or `onet_id`. If both `carotene_id` and `onet_id` are supplied,
+then the service first attempts to return work activities for the `carotene_id`. If none are 
+found then it falls back to using the `onet_id`. Company Work Activities is available at 
+`/core/company/workactivities`.
```
thomasmahoney: is that first sentence's logic out of date now? and maybe add "is available at [...] in the US, EU, and Staging regions" at the end (since not all APIs are available in all regions, mainly EU)?

CompanyWorkActivities.md
```diff
@@ -0,0 +1,54 @@
...
+  "company_id": "NCf01eccfd-78f5-43e5-98a1-9d6415ea4e7a",
+  "carotene_id": "13.308829999999999",
+  "onet_id": "53-6041.00"
+}
+```
+
+## Response Structure
+
+The response contains a list of work activities associated with the requested `company_id` and 
+`carotene_id` pair or `onet_id`.
```
thomasmahoney: this wording sounds out of date now, too (it sounds like it was written under the old paradigm where companyID wasn't required if you sent an onetID)

CompanyWorkActivities.md
```diff
@@ -0,0 +1,54 @@
...
+```json
+{
+    "data": {
+        "work_activities": [
+            "Assist with the administration of various employee benefits programs, including medical, dental, vision benefits",
+            "Facilitate weekly new-hire benefit orientations",
+            "Reconcile invoices from benefits vendors and providers"
+        ]
+    }
+}
```
thomasmahoney: I'm just now noticing that there is no indication in the response of whether the carotene or onet ID was used, if both were provided. maybe we should add that? or make `carotene_work_activities` and `onet_work_activities` nodes instead? (and then maybe it just fetches both if you pass both IDs? (ponder))

CompanyJobTitles.md
```diff
@@ -0,0 +1,58 @@
...
+
+- [Summary](#summary)
+- [Request structure](#request-structure)
+- [Response structure](#response-structure)
+- [Versioning](#versioning)
+
+## Summary
+
+The company Job Titles service provides a list of job titles (along with their ONet17 and Carotene 
+v3 classifications) that are personalized for a given company. Company Job Titles is
```
thomasmahoney: IMO the summary should just mention ONet and Carotene, and the taxonomy versions should be named in the Versioning section.

CompanyWorkActivities.md
thomasmahoney: also, we should probably call out which ONet/Carotene taxonomy version(s) this service supports (again, maybe the Versioning section is best for this). it would be a bad experience for the user to submit, say, a valid caroteneV1 ID and get no results because that ID doesn't exist in caroteneV3.

CompanyJobTitles.md
```diff
@@ -0,0 +1,58 @@
...
+```json
+{
+	"company_id": "NC6623058c-2e0b-4998-ac4a-72e478e91337"
+}
+```
+
+
+## Response Structure
+Response consist of a list of `job_titles` where each item contains a `job_title` string and 
+its associated ONet17 and Carotene v3 classification code.
```
thomasmahoney: again, I would avoid duplicating taxonomy version info here

CompanyWorkActivities.md
glpayson: I'm fine with having there be two separate nodes, but imo we shouldn't return the onet activities unless we have to. Janani was pretty explicit that onet should only be a fall back, so I think it is unlikely that the front end teams would want both. And if they do we can always add it without breaking the response contract.

CompanyWorkActivities.md
glpayson: Changed

CompanyWorkActivities.md
glpayson: Changed

CompanyWorkActivities.md
glpayson: I changed the response structure.

CompanyJobTitles.md
glpayson: Changed

CompanyJobTitles.md
glpayson: Changed

CompanyJobTitles.md
```diff
@@ -0,0 +1,64 @@
...
+
+```json
+{
+	"company_id": "NC6623058c-2e0b-4998-ac4a-72e478e91337",
+	"carotene_version": "carotenev3",
+	"onet_version": "onet17"
+}
+```
+
+Available taxonomies are `onet17` for `onet_version` and `carotenev3` for `carotene_version`.
```
thomasmahoney: this info is duplicated here - perhaps replace with something like "The ONet and Carotene versions currently supported are listed in the [Versioning](#Versioning) section of this document and are subject to change over time. Any removal of support for a taxonomy version will be implemented as a versioned change to the API itself."

[side note: should we do that second thing? I feel like probably?]

CompanyWorkActivities.md
```diff
@@ -0,0 +1,77 @@
...
+  "company_id": "NCf01eccfd-78f5-43e5-98a1-9d6415ea4e7a",
+  "carotene_id": "51.31",
+  "carotene_version": "carotenev3",
+  "onet_id": "53-6041.00",
+  "onet_version": "onet17"
+}
+
+```
+
+Available taxonomies are `onet17` for `onet_version` and `carotenev3` for `carotene_version`.
```
thomasmahoney: same comment applies here too

CompanyJobTitles.md
glpayson: Changed

CompanyWorkActivities.md
glpayson: Changed

PR #138, opened 2018-09-27
Update readme

thomasmahoney:


glpayson (approved):

README.md
```diff
@@ -1,42 +1,89 @@
...
 
-You will need OAuth core credentials to use these services. (*If you do not have these please contact API and Auth to request credentials*)
+Below, you will find general information pertaining to all DSAD APIs. **Please review this documentation, along with the relevant service-specific documentation, in full before starting to work with any DSAD APIs. By integrating with any of the APIs documented here, you indicate that you have reviewed the relevant documentation and agree to any obligations defined for our API clients.**
 
-We provide .net and java wrappers for our APIs, which can be found [here](https://github.com/cbdr/DataScienceServices). (Note: this repository may only be viewed by members of the cbdr GitHub organization.)
+### Contact Information
 
-Our general API versioning documentation can be found [here](Versioning.md).
+The Data Science Application Development team is the primary point of contact for API-related questions and concerns. The team can be reached at the following email:
```
msundlof, line 9: If you removed the email did you intend to remove this sentence?

msundlof: Or is it in the dsad.png? if so ignore this comment

README.md
```diff
@@ -1,42 +1,89 @@
...
+* **404 (Not Found)** : The API request targeted a route (or a version of a route) that does not exist.
+* **405 (Method Not Allowed)** : The API request specified an HTTP method that is not supported by the targeted API route and version.
+* **429 (Too Many Requests)** : The client_id associated with this request has sent too many requests to this API recently, and the API's performance is beginning to degrade as a result. Please wait a few minutes before resuming API calls.
+* **500 (Internal Server Error)** : The API request was accepted, but failed to complete successfully due to an internal error within the API. This error should not occur under normal circumstances; if you encounter this, please contact the DSAD team.
+* **502 (Bad Gateway)** : The API request was accepted, but failed to complete successfully due to an error which occurred with a downstream provider called during request processing. This error should not occur under normal circumstances; if you encounter this, please contact the DSAD team.
```
msundlof: I think it would be an upstream provider here. Also we don't want people calling us on any one 500, would it be better to state that we have a threshold of acceptable 500 depending the service and if you see consistent errors then contact us?

thomasmahoney: @msundlof yeah, you're right, upstream is correct here. I'll fix that.

I don't think we should generally accept any steady-state percentage of 500s. to me, any 500 from our API indicates a mistake on our part to be fixed -- either something that *should* be a 4xx, or a programming error that shouldn't happen at all. (note I'm referring to 500 specifically, not all 5xx's, though I also feel that we should work toward 0% steady-state 5xx's across our platform as well).

thomasmahoney: so I'd like to keep the "please contact us" verbiage for 500s, but I'm open to changing it for 502s. I'll bring it up in the team room and see what others think.

msundlof (approved):

PR #139, opened 2018-09-28
document zuul codes

thomasmahoney:


psequeirag (approved):
Thanks for pulling the link to the codes.

PR #140, opened 2018-10-12
Reranker update

feng0905:
Reranker request structure change:
1. location becomes an object, which contains "city", "state", "country", "latitude", and "longitude". 
2. major becomes majors.

### DIFF
--- a/CandidateReranking.md
+++ b/CandidateReranking.md
@@ -329,7 +329,7 @@ Each Job Title object only has a single title field associated as follows:
       "unnormalized_education_level": [
         "master's of science"
       ],
-      "major": ["Computer Science"],
+      "majors": ["Computer Science"],
       "city": "Atlanta",
       "state": "GA",
       "country": "US",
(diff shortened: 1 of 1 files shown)


CandidateReranking.md
```diff
@@ -118,7 +129,7 @@ A profile represents a candidate and is composed of data which, in relation to t
 | years_of_experience | Double | **FALSE** | Total years of experience the candidate has.
 | normalized_education_level | String[] | **FALSE** | Normalized Educations associated with the candidate i.e., Master's Degree, Bachelor's Degree etc.
 | unnormalized_education_level | String[] | **FALSE** | Unnormalized Educations associated with the candidate. i.e., masters degree, Master of Science, Masters etc.
-| major | String | **FALSE** | The major or field of study the candidate's degree pertains to.
+| majors | String[] | **FALSE** | The list of majors or fields of study the candidate's degree pertains to.
```
psequeirag: As this is Json, it's more correct to say "array of majors" rather than "list of majors".

psequeirag (approved):
Just a minor comment, but otherwise looks good to me.

CandidateReranking.md
feng0905: Yeah, array makes more sense. thanks.

glpayson (approved):

PR #143, opened 2018-10-29
Semantic autocomplete

msundlof:


SemanticAutocomplete.md
```diff
@@ -0,0 +1,127 @@
...
+
+- [Summary](#summary)
+- [Request](#request)
+- [Response](#response)
+- [Example](#example)
+- [Versioning](#versioning)
+
+
+## Summary
+The autocomplete service wraps the [Solr Suggester](https://lucene.apache.org/solr/guide/6_6/suggester.html) which is configured in DSAD's Semantic Search Solr boxes. Version 2 of the Semantic Autocomplete service is a complete rewrite of the original service and makes use of the latest FSTs provided to DSAD from Textkernel which are updated sparodically throughout the year. With the exception of some removed feilds from the response, this version of the service functions exactly like V1.
```
glpayson: I would say 'The Semantic Autocomplete service wraps ...' We have multiple autocomplete services after all.

glpayson: 'sporadically' but I wouldn't use that term. It has kind of a negative connotation.

SemanticAutocomplete.md
```diff
@@ -0,0 +1,127 @@
...
+
+
+## Request
+As per all DSAD services, this service accepts requests sent as either HTTP GET or HTTP POST. Access to semantic autocomplete is made available at:
```
glpayson: I'd say 'As with', 'As per' means in accordance with some rule or request.

SemanticAutocomplete.md
```diff
@@ -0,0 +1,128 @@
...
+
+- [Summary](#summary)
+- [Request](#request)
+- [Response](#response)
+- [Example](#example)
+- [Versioning](#versioning)
+
+
+## Summary
+The Semantic Autocomplete service wraps the [Solr Suggester](https://lucene.apache.org/solr/guide/6_6/suggester.html) which is configured in DSAD's Semantic Search Solr boxes. Version 2 of the Semantic Autocomplete service is a complete rewrite of the original service and makes use of the latest FSTs provided to DSAD from Textkernel which are updated with improvements throughout the year. With the exception of some removed feilds from the response, this version of the service functions exactly like V1.
```
glpayson: I feel like this first sentence is relatively low level and maybe not of much use to potential users unless they have an in depth knowledge of Solr and our Semantic Search service. I think it would be better to describe just at a high level what it does and/or why they would want to use it.

SemanticAutocomplete.md
```diff
@@ -0,0 +1,128 @@
...
+| Field | Type | Description |
+|-------|------|-------------|
+| query | String | Original query sent to Solr.
+| suggests | Suggestion[] | A json array of Suggestions broken up by type.
+
+### Suggestion
+
+| Field | Type | Description |
+|-------|------|-------------|
+| type | String | The type or cannonical form of the suggestion. Returned values include: *keyword, job_title, job_level, occupation, skill, school, company, location, freetext*
```
glpayson: I assume this is an exhaustive list so you might say instead "Possible returned values are: "

SemanticAutocomplete.md
```diff
@@ -0,0 +1,128 @@
...
+| values | value[] | A json array of values associated with the type.
+
+
+### Value
+A value object represents the surface forms associated with the type(cannonical) suggested for autocomplete. E.g., provided "e" as a query, under the `type` *job_title* you would see `names` like engineer, executive, sales executive, etc.
+
+| Field | Type | Description |
+|-------|------|-------------|
+| name | String | A name associated with the type classification.
+| popularity | Integer | An integer value representing the score of popularity indicating how often the value is used.
```
glpayson: I think this might be a bit better worded as "An integer value indicating how often the term appears in resumes or job postings."

SemanticAutocomplete.md
```diff
@@ -0,0 +1,128 @@
...
+
+#### Notable missing fields
+The older service maintained a timing object as part of the data field. This is no longer returned.
```
glpayson: Instead of 'The older service' I would say "Version 1 of this services returned a timing... This is no longer returned in Version 2."

SemanticAutocomplete.md
```diff
@@ -0,0 +1,128 @@
...
+- [Request](#request)
+- [Response](#response)
+- [Example](#example)
+- [Versioning](#versioning)
+
+
+## Summary
+The Semantic Autocomplete service wraps the [Solr Suggester](https://lucene.apache.org/solr/guide/6_6/suggester.html) which is configured in DSAD's Semantic Search Solr boxes. Version 2 of the Semantic Autocomplete service is a complete rewrite of the original service and makes use of the latest FSTs provided to DSAD from Textkernel which are updated with improvements throughout the year. With the exception of some removed feilds from the response, this version of the service functions exactly like V1.
+
+For ease of use DSAD autogenerates and publishes Java and C# SDKs for all of our services using swagger. For more information on usage please see our SDK repo [here](https://github.com/cbdr/dsad-sdks).
```
glpayson: I would maybe mention where they can get the SDKs (i.e. jfrog).

feng0905 (approved):
+1

SemanticAutocomplete.md
```diff
@@ -46,7 +46,7 @@ The response data is broken up into two parts: the original `query` and the list
 
 | Field | Type | Description |
 |-------|------|-------------|
-| type | String | The type or cannonical form of the suggestion. Returned values include: *keyword, job_title, job_level, occupation, skill, school, company, location, freetext*
+| type | String | The type or cannonical form of the suggestion. Possible returned values include: *keyword, job_title, job_level, occupation, skill, school, company, location, freetext*
```
glpayson: I still think we should say "values are" instead of "values include". Include implies there are other values that we have not listed here.

msundlof:
Lamar +1 on hipchat

PR #146, opened 2019-01-07
Documentation for JobToResume

akhilssct:


JobToResume.md
```diff
@@ -0,0 +1,69 @@
...
+_______
+
+- [Summary](#summary)
+- [Request structure](#request-structure)
+- [Response structure](#response-structure)
+- [Versioning](#versioning)
+
+## Summary
+
+JobToResume Recommend resumes based on job data. JobToResume is available at
```
glpayson: Typo here, and should probably be `The Job To Resume service recommends resumes... The Job To Resume service is available at ... '

JobToResume.md
```diff
@@ -0,0 +1,69 @@
...
+    "location": {"latitude": 40.7831, "longitude": -73.9712},
+    "carotene_version": "carotenev3"
+      }'
+```
+
+Currently only Carotene v3 is supported. See the documentation for [Job Title service](https://github.com/careerbuilder/DataScienceAPIDocumentation/blob/master/JobTitle.md)
+for more information about carotene classification and carotene IDs.
+
+## Response Structure
+The response consist of a list of `resumes` where each item contains a `id` string which is resumedid and `score` double which is cosine similarity score of resumedids which is a measure of relevancy of resumdids to the job data in request.
```
glpayson: First part should be 'where each item contains an `id` string with is a resumedid ...'

For the second part I think maybe we shouldn't go into detail about how the score is actually computed or what it is. I think we can just say something like 'and a `score` double which represents the how well a given resume matches the job data in the request.

glpayson (approved):

JobToResume.md
```diff
@@ -0,0 +1,69 @@
...
+    "location": {"latitude": 40.7831, "longitude": -73.9712},
+    "carotene_version": "carotenev3"
+      }'
+```
+
+Currently only Carotene v3 is supported. See the documentation for [Job Title service](https://github.com/careerbuilder/DataScienceAPIDocumentation/blob/master/JobTitle.md)
+for more information about carotene classification and carotene IDs.
+
+## Response Structure
+The response consist of a list of `resumes` where each item contains an `id` string which is a resumedid and `score` double which represents how well a given resume matches the job data in the request.
```
feng0905, line 49: I think it is better to change 'resumedid' to 'resume did' or 'resume_did' .

feng0905 (approved):
+1

JobToResume.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Request Structure
+Requests consist of:
+
+|                   |               |
+|-------------------|---------------|
+|`carotene_id`      | string        | 
+|`carotene_version` | string        |
+|`skills`           | List(Skill)   |
+|`location`         | Location      |
```
psequeirag: Don't we want to add a brief description of what each parameter does or is used for?

psequeirag: In this table, I mean.

psequeirag: Also, this is JSON, so there are no Lists; refer to them as Arrays instead.

psequeirag: And the Location object structure would probably be need to be detailed as well, unless we can relax and let them discover it fully from the swagger docs or the generated SDK. Thoughts, anyone?

psequeirag: On a second thought, scratch the comment on the Location: the example is probably already pretty clear on what to put in a location.

akhilssct: we are giving a sample curl request which will give user a clear idea about the fields,thats why I didn't explain.If you think we need explanation,I can add.Thoughts?

psequeirag: Yes, I think the descriptions may still be needed: are all parameters required? Are some of them optional, what happens if omitted? You already had to clarify what are the acceptable values for `carotene_version`, perhaps that sentence belonged to the description of the param.

JobToResume.md
```diff
@@ -0,0 +1,69 @@
...
+
+```
+curl -X POST \
+  https://api.careerbuilder.com/core/jobtoresume \
+  -H 'Accept: application/json;version=1.0' \
+  -H 'Authorization: <BEARER_TOKEN>' \
+  -H 'Content-Type: application/json' \
+  -d '{
+	"carotene_id": "11.0",
+	"skills": [{"skill_did": "KS120L16S3RLJ82VQ08H"}, {"skill_did": "KS120L96KMYTDJ48NRSH"}],
```
psequeirag, line 38: Wait, skill objects are just `skill_did`s? Or can they have more fields? In that case, why an array of objects and not an array of strings?

JobToResume.md
akhilssct: ok

psequeirag (approved):

JobToResume.md
akhilssct: That was Lamar's decision.:)

PR #149, opened 2019-02-21
Update documents for CNORM and CRS

feng0905:


CompanyNormalization.md
```diff
@@ -55,7 +55,7 @@ Following is an example JSON response body:
         "country": "US",
         "address": "200 N La Salle St # 1100",
         "naics_description": "Employment Placement Agencies",
-        "duns_number": "095301110",
+        "duns_number": "",
```
msundlof: I would mark this with a comment // deprecated also there is an additional duns_number on the part of the response below.

CompanyNormalization.md
```diff
@@ -27,12 +27,12 @@ The following parameters may be used in constructing a request to the service:
...
+The response for a search request returns a single `data` node which contains a `normalized_companies` array. These normalized companies are ordered by the confidence score (descending). Each normalized company has a `normalized_name` (string), a `naics_code` *([more info](http://www.census.gov/eos/www/naics/))* (string), a `naics_description` (string), a `duns_number`[<sup>[1]</sup>](1) (string), a `country` (string), a `state` (string), a `city` (string), a `postal_code` (string), a `website` (string), a `company_size` (int), an `id` (string), and a `confidence` (decimal). Confidence scores range from 0.0 to 1.0. A single master company will be returned.
```
msundlof: I don't think your link is working at least not when I tried it. And I would mention here that the string is always empty.

CompanyNormalization.md
```diff
@@ -92,3 +92,5 @@ Lookup responses are identical to search responses in structure, with the `maste
 The response from the Company Normalization Service is versioned with the current version being 1.0. The CompanyDepot data is also versioned, and we will always use the latest.
 
 Our general versioning strategy is available [here](/Versioning.md).
+
+[1] This number is now deprecated and set to empty string. It would be removed from later version.
```
msundlof: I would be more explicit and mention duns_number. And I would say "the string will always return empty". minor grammar. I think it would read better with: "It *will* be removed from later versions."

CandidateReranking.md
```diff
@@ -32,11 +32,11 @@ https://api.careerbuilder.com/core/search/candidatereranker.
 OAUTH credentials are **required**.
 
 A request is composed of 4 main parts:
-`reranker_config`, `source`, `query`, and `profiles`.
+`model_version`, `source`, `query`, and `profiles`.
 
 | Param    | Type | Required | Description |
 |----------|------|----------|-------------|
-| reranker_config | String | **TRUE** | Accepted Value: **RERANKER_V1** The `reranker_config` determines which learned model to apply to your incoming rerank request. Currently there is only one reranker model to call, "RERANKER_V1".
+| model_version | String | **TRUE** | Accepted Value: **X.Y** The `model_version` determines which learned model to apply to your incoming rerank request. Currently there is only two reranker models to call, **1.0** and **1.1**. The default value is 1.0.
```
msundlof: "Currently there *are* only two...."

msundlof (commented):
just a few little comments.

msundlof (approved):

PR #150, opened 2019-03-08
Vector Service Documentation

msundlof:
Documentation for the Vector Service.

### DIFF
--- a/VectorService.md
+++ b/VectorService.md
@@ -21,14 +21,14 @@ Requests can be sent in the form of GET or POST (form or json).
 
 ## Request
 
-A request can be broken out into two distinct types identified through a parameter called `vector_type`. Available  `vector_type`'s are: **carotene_skills** and **semantic_fingerprint**.
+A request can be broken out into two distinct types identified through a parameter called `vector_type`. Available  `vector_types` are: **carotene_skills** and **semantic_fingerprint**.
 
 ### Carotene Skills
 
 | Parameter  | Type | Required |  Description |
 |------------|------|----------|--------------|
 | carotene_id | String | True | A carotene ID as provided by our Job Classification services which uniquely identifies a job title.
-| carotene_version   | String | True | The version of carotene the ID is associated with.
+| carotene_version   | String | True | The version of carotene the ID is associated with. Currently the only accepted version is **CAROTENEV3**.
 | skills | List | False | A list of skill objects.
 
 Each skill object is made up of a unique number/letter identifier provided as a String.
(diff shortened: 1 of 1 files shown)


VectorService.md
```diff
@@ -0,0 +1,142 @@
...
+
+#### CaroteneSkills Only Request Sample
+```
+{
+  "carotene_id": "11.0",
+  "carotene_version": "CAROTENEV3",
+  "skills": [
+    {
+      "id": "KSH9TO"
+    }
```
psequeirag, line 48: Now that I pay attention to the request format. Was it not possible to pass Skill IDs as simply an array of strings instead of putting objects in the array?

VectorService.md
```diff
@@ -0,0 +1,142 @@
...
+This service provides vector representations for documents such as a Job or Resume, and for Carotene IDs with optional skills.
+
+
+The service is located at https://api.careerbuilder.com/core/document/vectors.  This service requires CareerBuilder OAuth 2.0 client credentials. For more information please see the authorization [documentation](/Readme.md#access).
+
+Requests can be sent in the form of GET or POST (form or json).
+
+## Request
+
+A request can be broken out into two distinct types identified through a parameter called `vector_type`. Available  `vector_type`'s are: **carotene_skills** and **semantic_fingerprint**.
```
psequeirag: The parameter name is actually in plural form: `vector_types`.

VectorService.md
```diff
@@ -0,0 +1,142 @@
...
+  "document_type": "job",
+  "vector_types": ["carotene_skills", "semantic_fingerprint"]
+}
+```
+
+
+## Response Structure
+
+All responses with an HTTP status of 200 will consist of a JSON object with a top-level "data" node. The response will contain each `vector_type` requested and an associated list of decimal numbers representing the computed vectors for the request.
+
```
akhilssct: But response will not contain both If I am requesting only one of the algorithms right. Do we need to mention it?

akhilssct (approved):

VectorService.md
```diff
@@ -0,0 +1,142 @@
...
+## Request
+
+A request can be broken out into two distinct types identified through a parameter called `vector_type`. Available  `vector_type`'s are: **carotene_skills** and **semantic_fingerprint**.
+
+### Carotene Skills
+
+| Parameter  | Type | Required |  Description |
+|------------|------|----------|--------------|
+| carotene_id | String | True | A carotene ID as provided by our Job Classification services which uniquely identifies a job title.
+| carotene_version   | String | True | The version of carotene the ID is associated with.
```
psequeirag: Are the valid values for carotene_version documented in some other place?

msundlof: ah they are not. I need to put that here. Good catch!

VectorService.md
msundlof: hmm I think we do this purposefully because it caused us some pain in another service. @glpayson do you recall?

msundlof: @psequeirag Lamar responded to me over the weekend, and the reason why we chose to do this is to allow more flexibility in the future should more fields i.e., name will be needed for vector calculations.

psequeirag (approved):
Good work, thank you!

PR #152, opened 2019-04-03
CompanyAddress

akhilssct:


CompanyAddress.md
```diff
@@ -7,124 +7,73 @@ _______
 - [Summary](#summary)
 - [Request structure](#request-structure)
 - [Response structure](#response-structure)
-- [Sample response](#sample-response)
 - [Versioning](#versioning)
 
 ## Summary
 
-The Company Address Service takes a request containing a company name, latitude and longitude and returns a response describing a normalized geographic location.
+The Company Address Service takes a request containing a company id, other company details as described in the below request structure and returns a response containing company Address. CompanyAddress is available at
```
glpayson: Casing on `containing company Address`

CompanyAddress.md
```diff
@@ -7,124 +7,73 @@ _______
...
+          "country": "US"
+      }'
+```
 
-&nbsp;
 
-## Sample Response
 
+## Response Structure
+All responses with an HTTP status of 200 will consist of a JSON object with a top-level "data" node. The response consists of company address and other location parameters like country,state,city,postal code,latitude,longitude and branch codes as show in the below sample response.
```
glpayson: need a space between the commas `country,state,city,postal code,latitude,longitude`

glpayson: Also could we not the latitude and longitude are optional on the response?

glpayson (approved):

psequeirag (approved):

CompanyAddress.md
```diff
@@ -7,124 +7,73 @@ _______
...
 - [Request structure](#request-structure)
 - [Response structure](#response-structure)
-- [Sample response](#sample-response)
 - [Versioning](#versioning)
 
 ## Summary
 
-The Company Address Service takes a request containing a company name, latitude and longitude and returns a response describing a normalized geographic location.
+The Company Address Service takes a request containing a company id, other company details as described in the below request structure and returns a response containing company address. CompanyAddress is available at 
+`/core/geography/companyaddress`.
```
msundlof: I would be more explicit with what the service is returning i.e. your not just returning  an address, your returning the address of the normalized company. I'm guessing because the response provides branch details that its not always the master company but can be a branch. Maybe it would be best to say that the service will return the nearest branch to the location and company provided in the request?

CompanyAddress.md
```diff
@@ -7,124 +7,73 @@ _______
...
+Requests consist of:
 
-&nbsp;
+| Field Name        | Type   | Required |
+|-------------------|--------|----------|
+|`company_id`      | string | True      |
+|`country` | string | False       |
+|`state` | string | False       |
+|`city` | string | False       |
+|`branch` | string | False       |
```
msundlof, line 27: Could you provide an example of what a branch string would look like?

CompanyAddress.md
```diff
@@ -7,124 +7,73 @@ _______
...
-| level | An integer between 1 and 5 inclusive specifying the hierarchal level at which this administrative area resides. In the US, states will be returned with a level of 1 and counties with a level of 2. **This property is always present and non-empty.** |
+## Request Structure
+Requests consist of:
 
-&nbsp;
+| Field Name        | Type   | Required |
+|-------------------|--------|----------|
+|`company_id`      | string | True      |
+|`country` | string | False       |
+|`state` | string | False       |
```
msundlof, line 25: Is this the full name or a 2 letter ISO identifier? Same with the country.

CompanyAddress.md
```diff
@@ -7,124 +7,73 @@ _______
...
-                    },
-                    "suggested_radius_miles": 0.10459088520608802
-                },
-                "partial_match": false
-            }
-        ],
-        "expiration_date": "2018-04-22"
-    }
+  "data": {
+    "id": "NC496cd8bb-93bf-4fe1-b9fc-4c966f4f95e5",
```
msundlof, line 56: Is this the same Id that the client provides in the request? If the client already has this info is it valuable to return it?

CompanyAddress.md
msundlof: I'd also like to know what effect on the results adding a branch name would have on the final results.

msundlof: Could we provide the full URL with host? If we had a sister company like Broadbean start using this they may not be familiar with what all careerbuilder service endpoints look like and would be kind of lost when we point them to this page. You could also link to the ReadMe which provides more comprehensive general information that all our services share.

CompanyAddress.md
```diff
@@ -7,124 +7,73 @@ _______
...
-
-Each element of the returned **admin_areas** array will be formatted as follows:
-
-|   |   |
-|---|---|
-| short_name | The abbreviated form of the administrative area's name. This will vary by country and administrative division level. In the US, states (admin_area level 1) will use the state abbreviation (e.g. "GA"), and counties will use the full county name (e.g. "Gwinnett County"). **This property is always present and non-empty.** |
-| long_name | The unabbreviated form of the administrative area's name. **This property is always present and non-empty.** |
-| level | An integer between 1 and 5 inclusive specifying the hierarchal level at which this administrative area resides. In the US, states will be returned with a level of 1 and counties with a level of 2. **This property is always present and non-empty.** |
+## Request Structure
+Requests consist of:
```
feng0905: Should be singular, right?

CompanyAddress.md
akhilssct: @glpayson : Need ur help here

CompanyAddress.md
akhilssct: True

CompanyAddress.md
akhilssct: I think from the sample request this is conveyed.

CompanyAddress.md
akhilssct: please see the sample curl request

CompanyAddress.md
akhilssct: so the nomalized name is of master company and  address will be that of branch if given ,otherwise the location information in the request should be exact match...if we get multiple matches will return empty response.

feng0905 (approved):
+1

CompanyAddress.md
glpayson: I think either is fine.

CompanyAddress.md
glpayson: I think it is okay. It is coming back from the lucene index, so we would need to specifically exclude it. Also, we do this sort of thing when doing `place_id` lookup with the geocoding service, so it isn't like this is without precedent:

```
[~] curl -X POST \
  'https://wwwtest.api.careerbuilder.com/core/geography/geocoding?culture=en&place_id=ChIJd8BlQ2BZwokRAFUEcm_qrcA' \
  -H 'Accept: application/json;version=1.1' \
  -H 'Authorization: Bearer tb19UUIzlq-3OFH1by2OEPa3wve0HR6bbAIpxPa3buHEfBcI5xjPBrjckKHliw_8HqGvhu0pwLCL74dY9BaoeBvZmr2QfNDWOCRdhU20qY_HlRPC6_Zs2wXNmCjlWpLOej_mL1gXQn4fJbVxvrzejrGSl3KHfD2DaZV6UMNe8DXy-9Q=='
{
  "data": {
    "results": [
      {
        "admin_areas": [
          {
            "long_name": "New York",
            "short_name": "NY",
            "level": 1
          },
          {
            "long_name": "Kings County",
            "short_name": "Kings County",
            "level": 2
          }
        ],
        "country": "United States",
        "country_code": "US",
        "formatted_address": "277 Bedford Ave, Brooklyn, NY 11211, USA",
        "latitude": 40.7142205,
        "longitude": -73.9612903,
        "location_type": "STREET_ADDRESS",
        "postal_code": "11211",
        "street_address": "277 Bedford Avenue",
        "sublocality": "Brooklyn",
        "place_id": "ChIJd8BlQ2BZwokRAFUEcm_qrcA",
        ...
```

msundlof (approved):

PR #153, opened 2019-05-28
Add JobRecommendationService.md

feng0905:


JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+# Summary
+
+Job Recommendation Service is part of the profile recommendations, which recommends similar and related jobs to a user based on the jobs he applied and user preferences. The recommendations are combined from FAISS indexes and GBR data, then filtered by the applied jobs.
+
+The endpoint for Job Recommendation Service is at https://api.careerbuilder.com/core/recommendations/job
+
+----------------------------
+# Request structure
+The service supports GET and POST requests to:
+https://api.careerbuilder.com/core/search/candidatereranker.
```
glpayson: I think this is left over copy paste?

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+    - [Compensation](#compensation)
+    - [CaroteneTitle](#carotenetitle)
+    - [CaroteneVersion](#caroteneversion)
+- [Response Structure](#response-structure)
+- [Versioning](#versioning)
+
+----------------------------
+# Summary
+
+Job Recommendation Service is part of the profile recommendations, which recommends similar and related jobs to a user based on the jobs he applied and user preferences. The recommendations are combined from FAISS indexes and GBR data, then filtered by the applied jobs.
```
glpayson: In general, we usually try to avoid giving implementation details about our services in our docs. To that end, I would remove the reference to FAISS and GBR. I also don't really think we need to mention profile recommendations. I would just say:

> Given a job ID and, optionally, a visitor ID, email, and a set of user preferences, the Job Recommendation Service provides recommendations of relevant and similar jobs.

JobRecommendationService.md
glpayson: Also, if you do fix this then it seems to make line 22 superfluous.

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+
+The endpoint for Job Recommendation Service is at https://api.careerbuilder.com/core/recommendations/job
+
+----------------------------
+# Request structure
+The service supports GET and POST requests to:
+https://api.careerbuilder.com/core/search/candidatereranker.
+
+OAUTH credentials are **required**.
```
glpayson: If you want to mention this then we should probably link to PS's OAuth docs: https://apimanagement.cbplatform.link/#/oauth/faq

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+
+A request is composed of 3 main parts:
+`job_data`, `user_data`, `user_preferences`.
+
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| `job_data` | [**JobData**](#jobdata) | **TRUE** | **JobData** contains a list of jobs which is needed for recommendations.
```
glpayson: should be 'which are needed'

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+| `user_data` | [**UserData**](#userdata) | **FALSE** | **UserData** contains visitor id and email address of the user, which is then used to retrieve applied jobs for filtering.
```
glpayson: I think this would be slightly better worded as:

> **UserData** contains the visitor id and email address for a given user. This data is used for filtering jobs that the user has already applied to from the final recommendations set.

JobRecommendationService.md
feng0905: you got me. Thanks.

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+| `user_preferences` | [**UserPreferences**](#userpreferences) | **FALSE** | **UserPreferences** indicates preferences of the user including location, compensation, job titles in carotenes and the associated varotene version.
```
glpayson: I would just leave this as:

>>> **UserPreferences** indicates preferences of the user including location, compensation, job titles.

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+
+### JobData
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| `source_jobs` | List<[**Job**](#job)> | **TRUE** | A list of **Job**.
+
+Note: currently only the first job in the list is used for recommendations.
```
glpayson, line 40: 👍

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+
+#### Job
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| `job_did` | String | **TRUE** | This String is required to start with `J` and have a length that is less than the given `MAX_LENGTH`, specifically 22 currently.
```
glpayson: I would just say that it must be a valid jobdid

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+
+### UserData
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| `visitor_id` | String | **FALSE** | This starts with `VX` and is required to be shorter than 50 (`MAX_LENGTH`) characters .
```
glpayson: Same thing here, I don't think there is much need to explain to people what the format of visitor id's is. I would just say it needs to be a valid visitor ID.

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+| `email_address` | String | **TRUE** | The email of the registered user.
+
+**UserData** contains two String parameters: `visitor_id` and `email_address`. However, currently only `email_address` is used to retrieve user applications.
```
glpayson: Hmm...I forgot about this. I would maybe just remove all references to visitor id from this document then. No sense in telling people about it when we don't support it yet.

feng0905: ok, I will remove it.

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+
+### UserPreferences
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| `location_filter` | [**PreferredLocation**](#preferredlocation) | **FALSE** | Preferered working location.
+| `compensation_filter` | [**Compensation**](#compensation) | **FALSE** | Expected compensation for a job application or a user.
+| `carotene_titles` | List<[**CaroteneTitle**](#carotenetitle)> | **FALSE**<sup>[1](#myfootnote1)</sup> | A list of carotene job titles that the user is interested.
+| `carotene_version` | [**CaroteneVersion**](#caroteneversion) | **FALSE**<sup>[1](#myfootnote1)</sup> | The email of the registered user.
```
glpayson: Copy paste mistake?

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+| `email_address` | String | **TRUE** | The email of the registered user.
+
+**UserData** contains two String parameters: `visitor_id` and `email_address`. However, currently only `email_address` is used to retrieve user applications.
+
+### UserPreferences
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| `location_filter` | [**PreferredLocation**](#preferredlocation) | **FALSE** | Preferered working location.
+| `compensation_filter` | [**Compensation**](#compensation) | **FALSE** | Expected compensation for a job application or a user.
+| `carotene_titles` | List<[**CaroteneTitle**](#carotenetitle)> | **FALSE**<sup>[1](#myfootnote1)</sup> | A list of carotene job titles that the user is interested.
```
glpayson: I would reword this as:

>>> A list of carotene job titles for the user.

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+| `carotene_version` | [**CaroteneVersion**](#caroteneversion) | **FALSE**<sup>[1](#myfootnote1)</sup> | The email of the registered user.
+
+**UserPreferences** provides additional constraints for recommendations. The service will rank recommendations from FAISS indexes and GBR data based on the location,  complesation and job titles. All the parameters are optional. But it has to be in valid format if any of them is provided.
+
+<a name="myfootnote1">1</a>: Carotene titles and version must be presented together or left blank, otherwise will be considered invalid.
```
glpayson: I would reword as:

`carotene_version` is required if `carotene_titles` is supplied.

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+
+#### PreferredLocation
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| `latitude` | double | **TRUE** | latitude of the location, range in `(-90, 90)`.
+| `longitude` | double | **TRUE** | longitude of the location, range in `(-180, 180)`.
+| `radius_in_miles` | float | **TRUE** | The redius of the location to specify the range of location filter, which has to be a positive number.
```
glpayson: I would reword this as:

>>> The radius surrounding the preferred location within which all recommended jobs must be located.

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+**PreferredLocation** uses coordinates along with the radius to indicate the preferred work location. All the parameters have to be present in valid values.
+
+#### Compensation
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| `type` | String | **TRUE** | Type of the compensation, currently only support `HOURLY` and `ANNUALLY`.
+| `amount` | double | **TRUE** | The total expected amount of the compensation, which is required to be a positive number.
+| `currency_code` | String | **TRUE** | The currency code of the expected compensation, identified by [ISO 4217](https://www.iso.org/iso-4217-currency-codes.html).
+
+**Compensation** is defined as composition of `type`, `amount` and `currency_code`, all of which are mandatory.
```
glpayson: Can we add a note that only USD is supported currently?

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+| `longitude` | double | **TRUE** | longitude of the location, range in `(-180, 180)`.
+| `radius_in_miles` | float | **TRUE** | The redius of the location to specify the range of location filter, which has to be a positive number.
+
+**PreferredLocation** uses coordinates along with the radius to indicate the preferred work location. All the parameters have to be present in valid values.
+
+#### Compensation
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| `type` | String | **TRUE** | Type of the compensation, currently only support `HOURLY` and `ANNUALLY`.
+| `amount` | double | **TRUE** | The total expected amount of the compensation, which is required to be a positive number.
```
glpayson: I would leave off the positive requirement. If people are passing in negative values on this field then they get the 400 they deserve imo.

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+| `currency_code` | String | **TRUE** | The currency code of the expected compensation, identified by [ISO 4217](https://www.iso.org/iso-4217-currency-codes.html).
+
+**Compensation** is defined as composition of `type`, `amount` and `currency_code`, all of which are mandatory.
+
+#### CaroteneTitle
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| `title` | String | **TRUE** | Job title from carotene taxonomy.
+| `id` | double | **TRUE** | The associated carotene id to the job title.
+Each **CaroteneTitle** is composed with `title` and associated `id`. The list of carotene titles can be found [here](https://github.com/cbdr/DataScienceAPITaxonomies/tree/master/JobTitle).
```
glpayson: Hmm instead of pointing them to the list, I would point them to Job Title Service instead. I'm afraid if we just link people to the list then it could mistakenly give the impression that they are on their own for picking a carotene title.

JobRecommendationService.md
```diff
@@ -0,0 +1,173 @@
...
+  "job_data": {
+      "source_jobs": [
+          {
+              "job_did": "J2W4YS61P1FX58XCNBT"
+          }
+      ]
+  },
+  "user_data": {
+      "visitor_id": "VX0E66C150DDE311E89FF2E5FA5B5A794816285H",
+      "email_address": "esat.sulejmani56@gmail.com"
```
glpayson: Please change this to a made up email address.

JobRecommendationService.md
glpayson: I think it might be a little better to put these summaries before their corresponding table rather than after them.

JobRecommendationService.md
feng0905: I used `Java.Util.Currency` class to check this, so it won't report report other currency as invalid now. Also, the currency and job titles currently are not used. Do we also need to remove them now?

JobRecommendationService.md
feng0905: Alright, let me remove this. 
Actually, this haven't been checked yet. so I created this pr https://github.com/cbdr/job-recommendation-service/pull/32.

JobRecommendationService.md
feng0905: Not really, the footnote is to clarify they must come together so I include both of them.

JobRecommendationService.md
glpayson: hmm yeah let's remove them if they aren't implemented yet.

akhilssct (approved):

JobRecommendationService.md
```diff
@@ -0,0 +1,131 @@
...
+    - [PreferredLocation](#preferredlocation)
+- [Response Structure](#response-structure)
+- [Versioning](#versioning)
+
+----------------------------
+# Summary
+
+Given a job ID and, optionally, an email, and a set of user preferences, the Job Recommendation Service provides recommendations of relevant and similar jobs.
+
+The endpoint for Job Recommendation Service is at https://api.careerbuilder.com/core/recommendations/job.
```
glpayson: I feel like we don't really need this line given lines 23 and 24 just below it.

JobRecommendationService.md
```diff
@@ -0,0 +1,131 @@
...
+
+----------------------------
+# Request structure
+The service supports GET and POST requests to:
+https://api.careerbuilder.com/core/recommendations/job.
+
+OAUTH credentials are **required** and the reference is provided [here](https://apimanagement.cbplatform.link/#/oauth/faq).
```
glpayson: Sorry, I should have been more specific maybe, but I would reinforce that OAuth is standardized across CB and not specific to us.

>>> OAuth credentials are **required**. Documentation for CareerBuilder authentication is provided [here](https://apimanagement.cbplatform.link/#/oauth/faq) by the Platform Software team.

JobRecommendationService.md
```diff
@@ -0,0 +1,131 @@
...
+
+A request is composed of 3 main parts:
+`job_data`, `user_data`, `user_preferences`.
+
+| Param    | Type | Required | Description |
+|----------|------|----------|-------------|
+| `job_data` | [**JobData**](#jobdata) | **TRUE** | **JobData** contains a list of jobs which are needed for recommendations.
+| `user_data` | [**UserData**](#userdata) | **FALSE** | **UserData** contains the  email address for a given user. This data is used for filtering jobs that the user has already applied to from the final recommendations set.
+| `user_preferences` | [**UserPreferences**](#userpreferences) | **FALSE** | **UserPreferences** indicates preferences of the user including location, compensation, job titles.
```
glpayson: Need an 'and' before job titles.

JobRecommendationService.md
glpayson: I wasn't referring to the footnote, I meant that it says 'The email of the registered user.' for carotene_version

JobRecommendationService.md
glpayson: I think you missed this one?

glpayson (approved):

PR #179, opened 2022-09-28
Update SalaryService.md

rehrenfeld:


SalaryService.md
```diff
@@ -30,6 +32,7 @@ Requests consist of:
 |`postal_code` | Optional | string | A postal code.|
 |`salary_period` | Yes | string | Possible values are "Year", "Hour". | 
 |`country` | Yes | string |An ISO-3166 country code.|
+|`hostiste` | Yes | string |A valid CareerBuilder hostsite.|
```
glpayson: typo in 'hostiste'. And, since we're in the neighborhood anyway, could use a space after the second to last | (same on the previous line that isn't part of the PR). 

Question: is hostsite parsed? If not then what happens when you request one that doesn't exist. If so then can we enumerate the valid values here?

ashraj10c1 (approved):

SalaryService.md
rehrenfeld: Fixed

glpayson (approved):

